PR #5, opened 2024-08-28
Odin Bindings - First Pass

nicbarker:
This is a draft in-progress PR containing clay bindings for Odin, will tidy up and write docs once it's in a complete state.


Dudejoe870:
Hey, I know this is a draft but I was looking through it and noticed you forgot to add `@(default_calling_convention = "c")` to the foreign block! Odin has a custom calling convention, so this could lead to potential issues!
(the docs that reference this are here https://odin-lang.org/docs/overview/#foreign-system)

laytan:
- I would do `Vector2 :: [2]c.float`, this allows taking advantage swizzling and array programming
- Same thing with `Color :: [4]c.float`
- You can take advantage of attributes to do away with all the wrappers you got:
```odin
@(link_prefix="Clay_")
foreign Clay {
    MinMemorySize :: proc() -> c.uint32_t ---
}
```
This will link to `Clay_MinMemorySize` but the symbol will be `MinMemorySize` in Odin.
- There is `@(private)` which you could use to not allow calling the private symbols at all, not sure if applicable here but good to know
- Where you do `@(deferred_none=blah)` you may want to `@(require_results, deferred_none=blah)` this will force the "handling" of the return value so `clay.Rectangle(...)` would not compile, that should hint the user to doing the scope thing with `if clay.Rectangle(...) {`, they could still do `_ = clay.Rectangle(...)` of course
- I would make `ClayString` -> `String` so you don't have `clay.ClayString(...)`

Thanks for doing this BTW! I am excited to use this.

Also, getting this working with Odin's WASM story should be pretty do-able since you don't use the stdlib / emscripten.

nicbarker:
@laytan Thanks so much for taking a look! I really appreciate it. Some great TIL's in there, I'll get them in ASAP.

nicbarker:
Bindings are working and I've tested it out with a raylib example and it works great!
Will get onto writing the docs and getting the example worked out a little better tomorrow.

nicbarker:
These are in a decently working state, I've replicated the clay website using Odin + Raylib:

https://github.com/user-attachments/assets/f94b7a55-3d14-460a-a4db-47302e69e883

Going to merge them now and work a bit more on the docs / finessing later.

Dudejoe870:
This looks great, awesome work! On both the bindings, and the library itself!

PR #8, opened 2024-08-30
support wasm in Odin bindings

laytan:
Adds a build of a wasm object file and a foreign import for it. Using wasm in Odin also adds the requirement of a `default_calling_convention` attribute, and requires globals to be in the fully global foreign scope `foreign {` instead of `foreign Clay {`. I also took the liberty to move the private declarations into their own foreign block and marked that entire block private instead of putting the attribute on all declarations.

I do want to note that this may be a different way of doing it vs your web examples. This way specifically works with the `vendor:wgpu` and `vendor:wasm/webgl` packages of Odin, instead of writing the rendering code in JS itself.

Also note that wasm `.o` files are cross-platform afaik.

I have successfully integrated clay into my webgpu renderer (both native and actually on the web).


nicbarker:
This looks super good to me, great work!

PR #12, opened 2024-08-31
[bindings/odin] Fix memory leak in Odin example

Dudejoe870:
Should fix an apparent memory leak (you allocate a clone of the text in the raylib renderer, however the temp_allocator was never freed)


nicbarker:
Thanks, this is a great visual example of how the temp allocator works!

nicbarker (approved):

PR #13, opened 2024-08-31
[bindings/odin] Improve Odin bindings

FourteenBrush:
- Replaces `c.uint32_t`, `c_uint64_t`, etc with their respective odin type
- Get rid of the need for `clay.MakeString` on `Text()`, possibly other procedures
- Get rid of `IDI`, now uses a default param on `ID`

Some other considerations:
- Maybe remove the need for `clay.ID` in components, and pass the id string instead?


bindings/odin/clay-odin/clay.odin
```diff
@@ -31,8 +31,8 @@ Dimensions :: struct {
 
 Arena :: struct {
     label:          String,
-    nextAllocation: c.uint64_t,
-    capacity:       c.uint64_t,
```
nicbarker, line 35: Thanks for all the work here - appreciate it!
I'd just like to get a bit of clarification here from the discord, because I'm a bit confused about this section in the docs:

https://odin-lang.org/news/binding-to-c/#types

Screenshot 2024-09-01 at 10 38 14 AM

nicbarker: I have just asked and got clarification - there's no functional difference for sized types < register width - so yep this change is great! Get to get rid of a whole bunch of casts which is awesome

bindings/odin/clay-odin/clay.odin
```diff
@@ -31,8 +31,8 @@ Dimensions :: struct {
...
+    nextAllocation: u64,
+    capacity:       u16,
```
nicbarker, line 35: Just a typo here
```suggestion
    capacity:       u64,
```

nicbarker:
> Maybe remove the need for clay.ID in components, and pass the id string instead?

It would certainly make things more ergonomic - but I have plans to turn the result of `clay.ID` into a richer datatype than `u32` in future, and it will be less painful in terms of breaking changes if I can just swap the return type and params underneath 🙂

nicbarker:
Oops, pressed the close button by mistake.

nicbarker:
I will merge this a little later today when I have time 👍

PR #16, opened 2024-09-13
Fix some Odin bindings

laytan:
- properly enforce that the measure text function is a "c" function
- make sure the abi matches for OpenScrollElement and OpenFloatingElement, when compiling this to wasm it would complain that the function signatures did not match because these return a pointer to `Clay_LayoutElement`, I opted to make them return a `rawptr` instead of binding the entire type


nicbarker:
Thanks for this, great catch!

PR #23, opened 2024-09-21
Use shallow clone on Raylib

SeverinDenisenko:
This change reduces the time of project configuration.

On my machine with decent internet connection, I got these results:
Without shallow clone: `cmake ..  24.75s user 6.55s system 54% cpu 57.140 total`
With shallow clone: `cmake ..  12.41s user 4.48s system 39% cpu 42.758 total`

Thank you for your time and effort!


nicbarker:
Hello, great catch! Of course we don't need a deep clone of raylib 🙂
Great to see such a significant speed up on first build.

PR #25, opened 2024-09-24
Initialize text measurement hashmap

bullno1:
The code measurement text actually expects the array to be zero-ed.


nicbarker (approved):
Great catch, on my system `malloc` etc always returns zeroed memory but that isn't guaranteed to be the case on every platform!

Issue #28, opened 2024-09-24
Suggestion: Use for loop for element macro


bullno1:
Instead of making the children code the last argument, the open/close calls can be made into a for loop:

```c
for (int i = (Clay__Open(...), 0), i < 1; ++i, Clay__Close()) {
    // children code
}
```

`CLAY_RECTANGLE` for example, can be defined as:

```c
#define CLAY_RECTANGLE(...) for (int i = (Clay__OpenRectangleElement(__VA_ARGS__), 0); i < 1;  ++i, Clay__CloseElementWithChildren())
```

It can be seen in action here: https://github.com/bullno1/cute-clay/blob/master/src/plugin_main.c#L76-L87
Definition: https://github.com/bullno1/cute-clay/blob/master/src/cute_clay.h#L17-L18

One advantage over the current method is that the children code is not smashed into one line by macro expansion.
It can be easily stepped over by a debugger.
Syntax highlighting or code completion would also be better since it's not inside a macro.

The disadvantage is that empty containers must end with `{}` or the next line will get roped into the loop.

nicbarker:
Hello and thanks for opening the issue, this is super interesting! I will do some local experimentation with this today and get back to you 😁

nicbarker:
This is genius! I'm a huge fan of the way it just "works" with the debugger, and prevents the macros being turned into one giant expansion. Some issues that need to be thought about:

- "Declaration shadows a local variable" warning because of the loop index variable. Not sure if there is a way around this one, but will have a play around and see.
- It's a bit of a silent footgun if the user forgets the `{}` and the subsequent element gets parented inside the previous one.

nicbarker:
Re: the first point, the following works ok, as long as you don't nest multiple elements on one line:

<img width="489" alt="Screenshot 2024-09-25 at 11 35 53 AM" src="https://github.com/user-attachments/assets/0a150ec5-e155-4985-8d12-e6ae58a4eac2">

nicbarker:
Ah - I see you solved it using a variable postfixed with the `__LINE__` of the macro. Seems like it's possible to use just a global latch here, something like this:

<img width="576" alt="Screenshot 2024-09-25 at 11 55 02 AM" src="https://github.com/user-attachments/assets/74b3372d-210f-4234-945a-288d6fbb2cc2">

Which will avoid having to define a whole bunch of local loop vars during layout creation 👍

nicbarker:
[...]

PR #30, opened 2024-09-28
[C] Convert element macros to use for() internally

nicbarker:
As per the wonderful suggestion by @bullno1 in https://github.com/nicbarker/clay/issues/28, this PR switches the core `CLAY_RECTANGLE` style C element macros to use a `for` construct internally rather than pasting in `children`. The primary reason for this is that rather than the entire UI hierarchy being folded into a single giant nested macro, this allows each element macro to be expanded separately, so that a debugger can step through it line by line. It also allows IDEs using auto complete / LSP to correctly display errors and warnings on code inside UI hierarchies, which significantly improves the developer experience for me personally.
Anecdotally it has also improved the performance of larger UI codebases in my personal editor, CLion.

The way this construct actually works internally is that `for` loops use the following form:

`for (definition; condition; increment) { body }`

With the key feature being that expressions are evaluated in this order:

`definition` -> `condition` -> `body` -> `increment` -> `condition` etc

The way the loop `body` is sandwiched between the `definition` and `increment` means that we can use the new definition of `CLAY__ELEMENT_INTERNAL` to open an element, then close it _after_ the body has executed exactly once:

```C
#define CLAY__ELEMENT_INTERNAL(open, close, ...) \
	for (\
		CLAY__ELEMENT_DEFINITION_LATCH = (open(__VA_ARGS__), 0); \
		CLAY__ELEMENT_DEFINITION_LATCH < 1; \
		++CLAY__ELEMENT_DEFINITION_LATCH, close() \
	)

#define CLAY_RECTANGLE(...) CLAY__ELEMENT_INTERNAL(Clay__OpenRectangleElement, Clay__CloseElementWithChildren, __VA_ARGS__)

// Example element
CLAY_RECTANGLE(CLAY_ID(), CLAY_LAYOUT(), CLAY_RECTANGLE_CONFIG()) {
    // ... loop body
}
```

With the above code execution order being:

- `Clay__OpenRectangleElement(id, layout, config)`
- `CLAY__ELEMENT_DEFINITION_LATCH = 0`
- `CLAY__ELEMENT_DEFINITION_LATCH < 1` (latch at this point is `0`, so true)
- `loop body` (child elements)
- `++CLAY__ELEMENT_DEFINITION_LATCH`
- `Clay__CloseElementWithChildren()`
- `CLAY__ELEMENT_DEFINITION_LATCH < 1` (latch at this point is `1`, so false)

As a result, we achieve the desired outcome of `open` -> `children` -> `close` with the loop body guaranteed to only ever execute once. The compiler should be able to infer this and remove the branch under `-O3`.

> [!WARNING]  
> **Unfortunately this is a small breaking API change for C Macros.**

The general migration strategy looks like this:

```C
// Old
CLAY_CONTAINER(id, layout, {
    // previously, children were nested inside parens ()
});

// New
CLAY_CONTAINER(id, layout) {
    // now, children are inside {} and CLAY_CONTAINER() is closed parens
} // <- note: no trailing semicolon ;

// Old
CLAY_CONTAINER(id, layout, {});

// New
CLAY_CONTAINER(id, layout) {}
```


clay.h
```diff
@@ -76,38 +76,30 @@
 
 #define CLAY_STRING(string) (Clay_String) { .length = CLAY__STRING_LENGTH(string), .chars = (string) }
 
+int CLAY__ELEMENT_DEFINITION_LATCH = 0;
```
bullno1, line 79: I think this should be static so that "clay.h" can be included in multiple compilation units.

nicbarker: Excellent point, and you've just reminded me to add a multi compilation unit test.

nicbarker: Fixed in https://github.com/nicbarker/clay/pull/30/commits/e3dacd93709688036624fa50a28ffa51824236e1

PR #31, opened 2024-09-28
C++20 Support

nicbarker:
This PR adds support for direct `#include clay.h` in C++20. The changes should be opaque to C and other languages. 

The main set of changes required were:

- Macro around designated initializers because C99 uses `(StructName) { members }` and C++20 uses `StructName { members }`
- Macro `__attribute__((__packed__))` and swap out for `enum : uint8_t` for 1 byte enums
- Reorder all designated initializer fields to be in declaration order, as required by C++20 standard (this was a pain)
- Swap out the C99 hack for detecting the alignment of a type for the supported `alignof` in C++20

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -1309,7 +1309,7 @@ typedef struct
     Clay__PointerInfoMouseDownState state;
 } Clay__PointerInfo;
 
-Clay__PointerInfo Clay__pointerInfo = (Clay__PointerInfo) { .position = {-1, -1} };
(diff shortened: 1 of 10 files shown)


clay.h
```diff
@@ -28,41 +28,47 @@
 #define CLAY_WASM_EXPORT(null)
 #endif
 
+#ifdef _MSC_VER
+#define CLAY_PACKED_ENUM
```
mikejsavage, line 32: You make this `: uint8_t` instead of a noop in C++

nicbarker: Nice idea, have put it in 👍

PR #33, opened 2024-10-02
Fix of a crash when mapItem is null

SogoCZE:
In some situations `mapItem` in `Clay_SetPointerState` can end up being null which results in to crash (because `elementId` is being accessed on null `mapItem`).

An example of this behaviour can be seen when you have an empty layout with only one text:
```c
Clay_BeginLayout();

CLAY_TEXT(
    CLAY_ID("HelloWorldText"), 
    CLAY_STRING("Hello World!"),
    CLAY_TEXT_CONFIG(
        .fontId = FONT_ID_BODY_24, 
        .fontSize = 24,
         .textColor = {255, 0, 0, 255}
    )
);
                             
return Clay_EndLayout();
```

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -3093,7 +3093,7 @@ void Clay_SetPointerState(Clay_Vector2 position, bool isPointerDown) {
             Clay_LayoutElementHashMapItem *mapItem = Clay__GetHashMapItem(currentElement->id); // TODO I wish there was a way around this, maybe the fact that it's essentially a binary tree limits the cost, have to measure
             if ((mapItem && Clay__PointIsInsideRect(position, mapItem->boundingBox)) || (!mapItem && Clay__PointIsInsideRect(position, CLAY__INIT(Clay_BoundingBox) {0,0, currentElement->dimensions.width, currentElement->dimensions.height}))) {
             	if (mapItem) {
-            		Clay__ElementIdArray_Add(&Clay__pointerOverIds, mapItem->elementId);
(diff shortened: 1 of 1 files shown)


nicbarker:
Hello, thanks for reporting this one! Taking a look now 👀

clay.h
```diff
@@ -3092,7 +3092,10 @@ void Clay_SetPointerState(Clay_Vector2 position, bool isPointerDown) {
             Clay_LayoutElement *currentElement = Clay_LayoutElementArray_Get(&Clay__layoutElements, Clay__int32_tArray_Get(&dfsBuffer, (int)dfsBuffer.length - 1));
             Clay_LayoutElementHashMapItem *mapItem = Clay__GetHashMapItem(currentElement->id); // TODO I wish there was a way around this, maybe the fact that it's essentially a binary tree limits the cost, have to measure
             if ((mapItem && Clay__PointIsInsideRect(position, mapItem->boundingBox)) || (!mapItem && Clay__PointIsInsideRect(position, CLAY__INIT(Clay_BoundingBox) {0,0, currentElement->dimensions.width, currentElement->dimensions.height}))) {
```
nicbarker, line 3094:
```suggestion
            if (mapItem && Clay__PointIsInsideRect(position, mapItem->boundingBox)) {
```

nicbarker:
Seems like the entire second half of that if statement was just an incorrect case. Must have been left over from a refactor or something. Nice catch!

PR #36, opened 2024-10-04
[Bug] Fix of broken text NEWLINES wrap mode

SogoCZE:
Wrap mode `NEWLINES` was skipped all the time because of this if condition: 
```c
// Short circuit all wrap calculations if wrap mode is none
if (textConfig->wrapMode == CLAY_TEXT_WRAP_NONE || (containerElement->dimensions.width == textElementData->preferredDimensions.width)) {
```
The second part of the if was always true for text with wrapping mode `NEWLINES`. I added another condition that skips this width check in the case of `NEWLINES`.

Automatic word wrapping (works fine):
image

Manual line breaks (`\n`) before the fix:
image

Manual line breaks (`\n`) after the fix:
image


nicbarker (approved):
Tested it out and looks good to me!

PR #41, opened 2024-10-11
Made casting more explicit for better compatibility with different compilers

johan0A:
Made casting more explicit for better compatibility with different compilers. For instance zig cc was producing errors.  And zig cc relies on some version of clang witch is probably also affected.


nicbarker:
@johan0A Thanks a bunch for fixing these up, I don't generally test with zig cc so wouldn't have caught these 🙂
I will see if I can get the github CI to run a compile using zig cc, so I can avoid breaking these compilers in future.

johan0A:
@nicbarker That would be very cool :D Though to be clear clay.h does compile without errors using zig cc but using the affected functions produces illegal instruction errors.

nicbarker:
@johan0A ah fascinating, good to know thanks!

PR #42, opened 2024-10-14
Fix scroll on mouse down being overriden by touch scroll

richardhozak:
Fixes unexpected behavior when clicking on scrollbar with mouse, the view would scroll the opposite direction as it would with touch scroll.

This behavior is unexpected with mouse.

The bug was probably caused by Clay_SetPointerState overriding scrollbarData and using the pointer state as scroll information for the container behind the scrollbar instead using the scrollbarData.

This fixes that behavior by not passing mouse down, when mouse down is captured by scrollbar.


nicbarker:
@richardhozak Hello, and nice catch! Thanks for the fix 🙂

PR #45, opened 2024-10-29
Add D example

bangbangsheshotmedown:
Hello,

I love your library, I use C and D, so I figured i'd share this example since you seems interested in providing bindings to other languages

D is a wonderful language, it also is a C compiler, so one can import C files directly, without needing to create bindings

I know D can understand C's macros, but I couldn't figure out a way to use them, i'll revisit that later

My code is not perfect, but I think this is a good starting point, maybe other D users can come and help improving it, D has fantastic template/mixin features, so I'm pretty sure this can be made much better.

Using your library from D is as simple as this:

`dmd main.d clay.c`

### DIFF
--- a/bindings/D/README.md
+++ b/bindings/D/README.md
@@ -1,0 +1,5 @@
+### D Language Example
+
+```
+dmd main.d clay.c
+```
--- a/bindings/D/clay.c
+++ b/bindings/D/clay.c
@@ -1,0 +1,2 @@
+#define CLAY_IMPLEMENTATION
+#include "../../clay.h"
--- a/bindings/D/main.d
+++ b/bindings/D/main.d
@@ -1,0 +1,88 @@
+import clay;
+
+import core.stdc.stdlib;
+
+__gshared:
+
+Clay_LayoutConfig layoutElement = { padding: {5} };
+
+extern(C) void main()
+{
+    ulong totalMemorySize = Clay_MinMemorySize();
+    Clay_Arena clayMemory = {
+        label: str("Clay Memory Arena"),
+        capacity: totalMemorySize,
+        memory: cast(char*)malloc(totalMemorySize)
+    };
+
+    Clay_Initialize(clayMemory, Clay_Dimensions(1024,768));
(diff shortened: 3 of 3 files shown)


nicbarker:
Hello, welcome and thanks for opening this PR! 🙂 It would be great to get some nice D bindings available, I've never used the language myself but have heard good things about it.

Would you be able to put together some example code to show how these bindings would be used?

bangbangsheshotmedown:
Thanks the plan, to port the current RayLib backend example with WASM support as well

nicbarker:
Awesome, great to hear!

cyrusmsk:
Hey @bangbangsheshotmedown 
If you are still interested there are D bindings now. We have SDL3 renderer and some code for raylib renderer as well (but raylib need a bit more work to prepare a proper version of the renderer).
If you are still interested in this usage feel free to use and contribute. The project author is pretty responsive (he merged my sdl3 renderer implementation)
And we are discussing library improvements in D discord as well.

https://github.com/nicbarker/clay/pull/607

bangbangsheshotmedown:
@cyrusmsk you don't need binding, D can import the .c file directly

bangbangsheshotmedown:
https://github.com/zkxjzmswkwl/clayd/blob/main/source/clayd/renderer/clay_sdl_renderer.d#L63

Why you malloc there? this is very bad

cyrusmsk:
> @cyrusmsk you don't need binding, D can import the .c file directly

I'm pretty aware of how ImportC is working.
But raw importC have their own drawbacks.. And it is not always work well with macro-heavy libraries.
Also, having API closer to D is nice

cyrusmsk:
> https://github.com/zkxjzmswkwl/clayd/blob/main/source/clayd/renderer/clay_sdl_renderer.d#L63
> 
> Why you malloc there? this is very bad

as D doesn't have VLA and to be able to support -betterC ..
But the library is still progressing. Maybe `alloca` could be used instead.
Feel free to improve the code =)

PR #47, opened 2024-11-14
docs: remove some inconsistencies with current API

27justin:
As a newcomer to the library, I noticed that the examples in the README are not 100% consistent with the actual API anymore.
I've updated the README to reflect the new API where I've noticed the inconsistencies.


nicbarker:
Hello, and thank you so much for picking these up, I really appreciate it 😁

nicbarker (approved):

PR #48, opened 2024-11-17
Include new cairo renderer

27justin:
Hello,

this PR adds a simple cairo renderer as well as a corresponding example to clay, this allows rendering directly into PNGs, PDFs, etc.
I've got it to a point where I believe it to be 'good enough' for now, it notably does not include a proper text shaping engine, instead relying on cairo's toy text API to measure text, etc.
This is flaky for some fonts, though I tweaked it that the result is acceptable for most fonts I tested it against.

Here's a preview of the PDF the example emits:
[output.pdf](https://github.com/user-attachments/files/17790796/output.pdf)


nicbarker:
Thanks for all the awesome work here 🎉 just taking a look now!

PR #49, opened 2024-11-17
Forward declare Clay__OpenTextElement

bullno1:
`CLAY_TEXT` won't work without this.


nicbarker (approved):
Thanks, mistake that this one was missed. Will add a test for it 👍

PR #50, opened 2024-11-18
Fix local id calculation

bullno1:
The `Clay_LayoutElement` type is not exposed so a function: `uint32_t Clay__GetParentElementId()` is needed.


nicbarker:
Thanks, and sorry for taking so long to merge this 😅 It got lost in the todo list!

PR #52, opened 2024-11-30
Improve overflow handling / CLAY_MAX_ELEMENT_COUNT exceeded

nicbarker:
This PR greatly improves the story around capacity overflows in various internal data structures.

Currently there are still some unsolved (but not unsolvable) problems:

- Text will not be rendered if `CLAY_MEASURE_TEXT_CACHE_SIZE` is exceeded.
- The error screen for exceeding `CLAY_MAX_ELEMENT_COUNT` is quite severe and non configurable:

Screenshot 2024-11-30 at 7 42 52 pm

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -3405,6 +3405,9 @@ void Clay_SetLayoutDimensions(Clay_Dimensions dimensions) {
 
 CLAY_WASM_EXPORT("Clay_SetPointerState")
 void Clay_SetPointerState(Clay_Vector2 position, bool isPointerDown) {
+    if (Clay__debugMaxElementsLatch) {
+        return;
+    }
     Clay__pointerInfo.position = position;
     Clay__pointerOverIds.length = 0;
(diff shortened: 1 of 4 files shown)


bullno1:
What about user-provided hook that defaults to assert/debugbreak?

Although tbh, if you reach that point, there's not much the program can do.

What does Clay do in that case though? Does it stop rendering new elements or just stop rendering altogether?

nicbarker:
@bullno1 Yeah it's a tricky situation, because if Clay runs out of allocated memory, we obviously don't want to just outright crash, and we _certainly_ don't want memory corruption, but at the same time the exhaustion is almost always going to happen half way through a layout declaration, which means the user is going to keep calling clay functions (potentially hundreds or thousands of times) _after_ we've run out of memory.

At the moment it just refuses to render and emits one render command with that warning message saying "run out of capacity", and I'm going to follow up with a PR that allows you to switch that off so it never happens even if you run out of memory in production.

PR #55, opened 2024-12-18
Fix: moved CLAY__MIN and CLAY__MAX to public macros

OleksiiBulba:
As raylib renderer uses CLAY__MIN and CLAY__MAX, I suggest moving those marcos to macros public definitions. Otherwise, it's impossible to compile raylib renderer without `#define CLAY_IMPLEMENTATION`.


nicbarker:
Good catch @OleksiiBulba, thanks for this one!

PR #65, opened 2024-12-20
Extract Clay Renderer API into a separate header file

OleksiiBulba:
In order to not depend on the specific renderer I suggest extracting renderer functions into its own header api file. Here what I did:

- Extracted clay_renderer_raylib.h;
- Modified Raylib and Cairo renderers to follow renderer api;

For Clay_Renderer_Initialize I've created an opaque struct type Clay_Renderer_Data so each renderer can populate its own config: for Raylib it's just window dimensions, title, and flags, for Cairo it's cairo_t type.

I'm not sure if it's ok to modify Cairo renderer, as it has own License notice.


nicbarker:
Hello, apologies for the delay, and thank you for this work! This is a great initiative, but it will be important to make sure we really get it right, so I'll leave this one open until we get a couple more renderers into the codebase and we can really see what the shared API surface will look like 🙂

PR #78, opened 2024-12-22
fix: move internal types to stdint specific. 

mrneo240:
Match Clay_RenderCommandArray_Get protoypes

This PR helps when using Clay on platforms that aren't x86_64.


nicbarker:
Great work, I will look these over an merge in a few hours 👍

nicbarker:
Looks good to me, thanks!

PR #81, opened 2024-12-22
[Core] Fix a couple of standards-compliance issues with C99

FintasticMan:
Doesn't fix the `##__VA_ARGS__` warnings, as they aren't easily solved with pure C99.

This only fixes the issues in the public code, more needs to be done to do the same for the implementation.

I guess this does cause an API change in the form of the `Clay_SizingAxis` struct being changed slightly, but I think everyone should be using the macros so it shouldn't matter that much.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -69,15 +69,15 @@
 #define CLAY__STRUCT_0_ARGS() {}
 #define CLAY__STRUCT_OVERRIDE(_0, _1, NAME, ...) NAME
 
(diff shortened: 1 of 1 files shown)


nicbarker:
Hello, thanks for the work! Just quickly, could you walk me through what the compliance issue is for `sizeMinMax` as it stands? 🙂

FintasticMan:
Of course! Anonymous unions (i.e. unions in structs or other unions without a name) were only added in C11. See https://en.cppreference.com/w/c/language/union.

FintasticMan:
@nicbarker I've just updated to the latest main branch, as there were some conflicts.

nicbarker:
@FintasticMan thanks for the work here!

PR #105, opened 2024-12-25
[Core] Implement Error Handler / Callback

nicbarker:
This PR changes `Clay_Initialize` to require that the user provide a function that Clay can call when it encounters an error.
The current error types include:

```C
typedef enum {
    CLAY_ERROR_TYPE_TEXT_MEASUREMENT_FUNCTION_NOT_PROVIDED,
    CLAY_ERROR_TYPE_ARENA_CAPACITY_EXCEEDED,
    CLAY_ERROR_TYPE_ELEMENTS_CAPACITY_EXCEEDED,
    CLAY_ERROR_TYPE_TEXT_MEASUREMENT_CAPACITY_EXCEEDED,
    CLAY_ERROR_TYPE_DUPLICATE_ID,
    CLAY_ERROR_TYPE_FLOATING_CONTAINER_PARENT_NOT_FOUND,
    CLAY_ERROR_TYPE_INTERNAL_ERROR,
} Clay_ErrorType;
```

This is part of work to allow the user to handle the case where clay runs out of space in its internal arena, and needs to reallocate with more memory.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -3180,7 +3180,9 @@ void Clay__RenderDebugView() {
             break;
         }
     }
-    int32_t highlightedRow = (int32_t)((Clay__pointerInfo.position.y - scrollYOffset) / (float)CLAY__DEBUGVIEW_ROW_HEIGHT) - 1;
+    int32_t highlightedRow = Clay__pointerInfo.position.y < Clay__layoutDimensions.height - 300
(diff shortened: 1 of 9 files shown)


clay.h
```diff
@@ -144,9 +144,8 @@ typedef struct
 } Clay__StringArray;
 
 typedef struct {
-    Clay_String label;
-    uint64_t nextAllocation;
-    uint64_t capacity;
+    uintptr_t nextAllocation;
+    size_t capacity;
```
nicbarker, line 148: Small types cleanup here. Should be transparent

appe12341 (approved):

PR #107, opened 2024-12-25
[Renderers/SDL3] Adds an example using SDL3 as a renderer

LiquidityC:
This example is rather basic but it does provide an example how one could setup a rendering loop between clay and SDL3.

Although SDL3 is in its infancy and doesn't have an official stable release the API has been locked so there shouldn't be any code adaptations required for this example if one was to update to a more recent SDL3 release in the future.


nicbarker:
Awesome, thank you for the contribution! I will take a proper look at this over the next week or so 👍

examples/sdl3_renderer/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
+cmake_minimum_required(VERSION 3.27)
+
+# Project setup
+project(clay_sdl3_renderer C)
+set(CMAKE_C_FLAGS_DEBUG "-g -Wall -Werror")
```
SiebrenW, line 5: Small remark here, which was already present in the other examples: you're replacing the flags here instead of adding them. Ideally, you'd write something like

```cmake
set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -g -Wall -Werror")
```

examples/sdl3_renderer/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+FetchContent_Declare(
+	SDL
+	GIT_REPOSITORY https://github.com/libsdl-org/SDL.git
+	GIT_TAG preview-3.1.6
+	GIT_SHALLOW TRUE
+	GIT_PROGRESS TRUE
+)
+message(STATUS "Using SDL via FetchContent")
+FetchContent_MakeAvailable(SDL)
+set_property(DIRECTORY "${CMAKE_CURRENT_BINARY_DIR}/_deps/sdl-src" PROPERTY EXCLUDE_FROM_ALL TRUE)
```
SiebrenW, line 21: You probably want to use `<lowercaseName>_SOURCE_DIR` instead of `_debs/sdl-src` https://cmake.org/cmake/help/latest/module/FetchContent.html

examples/sdl3_renderer/CMakeLists.txt
LiquidityC: Done

examples/sdl3_renderer/CMakeLists.txt
```diff
@@ -0,0 +1,47 @@
...
+
+# Download SDL_ttf
+FetchContent_Declare(
+	SDL_ttf
+	GIT_REPOSITORY https://github.com/libsdl-org/SDL_ttf.git
+	GIT_TAG 40219a6
```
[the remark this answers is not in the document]
LiquidityC: I'll swap it to main for now. Once there's an official stable SDL3 release i assume there'll be a related tag set in SDL_ttf. I think the risk of breakage should be rather low at this point even if we do point out main.

LiquidityC:
> Tested as working on macOS 15.2 -- thanks for this! I appreciate that you took the time to wrap in SDL_ttf. I also like that you used the new callbacks method for integration. I'm all-in on SDL3 so it's nice to see others adopting it for new, useful frameworks like Clay.
> 
> I left one comment on the SDL_ttf version, above. Otherwise, the demo works a treat.
> 
> My thoughts:
> 
> * Do you think it would be good to split this into a renderer that could be shared?
> * Should this be using dynamic libraries by default?
> 
> I'd also love to see the demo essentially replicate the Raylib scroller, which would involve adding SDL_image to the build. (I'm one of those weirdos who thinks SDL_image and SDL_ttf should be part of the base library, though. 🤣 )

Nice that it worked.

Regarding extensions. There is one caveat with the builtin renderer in SDL. There's no ability to render circles out of the box. I didn't look into the "border" rendering commands from clay. But my best guess is that there might be some issues with this and the `border radius`. This will require bespoke solutions per project that does chose to utilize clay.

For me that's a reason to avoid providing a default SDL renderer within clay and instead leave that to the user to solve for their particular case.

Granted I'm not the maintainer of the repo and if there's an interest in extending this renderer to a provided solution I'm open to help out.

examples/sdl3_renderer/CMakeLists.txt
LiquidityC: I did not know that this variable existed. Thanks. I'll need to revisit some of my projects now I think :smile:

dfyx:
You can also pin commit IDs when no branch or tag is available. I'd recommend doing that and adding a comment like "main as of 2025-01-10" for future reference.

dfyx:
For circles and rounded rectangles, it might be worth borrowing some code from SDL_gfx: https://www.ferzkopp.net/Software/SDL2_gfx/Docs/html/_s_d_l2__gfx_primitives_8h.html

PR #111, opened 2024-12-26
[Core] Access bounding boxes

hexmaster111:
I added an API method `Clay_GetElementLocationData(Clay_ElementId)`

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -393,6 +393,14 @@ typedef struct
     bool found;
 } Clay_ScrollContainerData;
 
+typedef struct
+{
+    Clay_BoundingBox elementLocation;
+
+    // Indicates whether an actual Element matched the provided ID or if the default struct was returned.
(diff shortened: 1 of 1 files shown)


bullno1:
Knowing the bound also helps with handling interaction.
For example, positioning a cursor when a textbox is clicked.

hexmaster111:
> Knowing the bound also helps with handling interaction.
> For example, positioning a cursor when a textbox is clicked.

I actually used it for just that with my sliders !

https://github.com/hexmaster111/clay_demo_app/blob/main/src%2Fslip.c#L71

emoon:
It would be great if this feature could be added

nicbarker:
Apologies, I accidentally did something janky with git and this one. Will merge a second commit with the changes with you as co-author, thanks for the work on this one! 🙂

Issue #75, opened 2024-12-21
Minor: Clay_Color range is unintuitive


mikejsavage:
Clay_Color is defined as 4 floats, so you would expect each component to range from 0-1, but the inspector hardcodes colours with components in 0-255 as if it was 4x uint8.

Besides the inspector clay doesn't seem to care what you put in Clay_Color so it's not a big deal, but it would still be nice to clear this up.

nicbarker:
@mikejsavage I totally agree, the inspector hardcoding colour values was a mistake. I wanted to make sure the data type was flexible enough to support formats that use 4x floats in 0-1 range, but they are 0-255 simply because that's what raylib asks for.

Perhaps the correct solution is that the user can optionally provide a function pointer that transforms Clay_Color -> Clay_Color that the debugger calls to transform it's internal colours into a format that the user decides on.

DiThi:
I think it would be better and more consistent to use colors 0 to 1 in linear space, and then leave it to the specific implementation (like raylib in this case) to convert them to uint8 in sRGB space or whatever. In my case I would use the colors in linear space directly, only converting everything to sRGB as the last step on the shader (or not even that if the target supports sRGB framebuffers). Also transitions and gradients look way better when interpolated in linear space.

To be compatible with existing code, you can auto detect the mode (0-255 or 0.0-1.0) if there's an alpha element over 1 and issue a deprecation warning.

leidegre:
I ran into this issues as well, it's great that it just passes whatever you provide to the backend and it can do what it wants. That way there's no extra work. I'm using a custom Direct2D backend and it is using color values in the range 0,1. This works, the only trick I had to employ was to convert the debug color values. So far, this is the only component that made an assumption about what color to use but that is easy to fix yourself.

```c
#define CLAY_COLOR_NORMALIZE(c,n) (Clay_Color){c.r/n,c.g/n,c.b/n,c.a/n}

CLAY__DEBUGVIEW_COLOR_1=CLAY_COLOR_NORMALIZE(CLAY__DEBUGVIEW_COLOR_1,255.0f);
CLAY__DEBUGVIEW_COLOR_2=CLAY_COLOR_NORMALIZE(CLAY__DEBUGVIEW_COLOR_2,255.0f);
CLAY__DEBUGVIEW_COLOR_3=CLAY_COLOR_NORMALIZE(CLAY__DEBUGVIEW_COLOR_3,255.0f);
[...]

PR #112, opened 2024-12-26
Make color type configurable (#75)

matevzmihalic:
Allow users to define `Clay_Color` type which would allow us to use the same type in clay and in renderer which would omit the currently needed conversions.
I've also added `CLAY_COLOR_RGB` and `CLAY_COLOR_RGBA` that should be used instead of hard coded color structs.

To use clay with grayscale renderer you would need to define this:
```
#define CLAY_COLOR uint8_t
#define CLAY_COLOR_RGBA(r, g, b, a) (uint8_t)roundf(0.2126 * (float)r + 0.7152 * (float)g + 0.0722 * (float)b)
```

If you think this is an ok solution we also need to change two more things in debug view:
* changing color for highlighted row
* outputing color string (currently in `{r: 0, g: 0, b: 0 a: 0}` format)

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -157,9 +157,21 @@ typedef struct {
     float x, y;
 } Clay_Vector2;
 
-typedef struct {
-    float r, g, b, a;
-} Clay_Color;
+#ifndef CLAY_COLOR
+#define CLAY_COLOR struct { \
+    float r, g, b, a; \
+}
+#endif
+
+#ifndef CLAY_COLOR_RGB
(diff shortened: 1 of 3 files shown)


nicbarker:
Thanks for opening this one, I will take a look today or tomorrow 🙂

nicbarker:
We should also do this for `Clay_String` I think, most users I suspect will have their own string container struct and it would be nice to avoid duplication boilerplate

PR #115, opened 2024-12-27
[Renderers/SDL2] Create initial SDL2 renderer

nicbarker:
This PR creates an SDL2 renderer and corresponding example.

In the process of getting SDL2 running I realised we aren't outputting the correct dimensions for wrapped lines of text. That has been fixed here.


ColleagueRiley:
@nicbarker I don't think allocating a texture/surface each time the user wants to draw text is a good idea...

Also I don't get the point of cloning the string, especially with calloc.

Issue #103, opened 2024-12-24
[Documentation] Better code commenting of macro magic


VisenDev:
I'm having a difficult time figuring out what exactly these lines are supposed to do, and I'd imagine i'm not the only one. Given that the `CLAY` macro is crucial to build anything with this library, it would be helpful having some more explanations of what these macros actually do. I've read thru them a few times and I'm still a bit lost.

https://github.com/nicbarker/clay/blob/6a9b723dcce154347452491559b34da528cd657f/clay.h#L95-L113

nicbarker:
Hello 👋 
Agree, there is some real trickery going on here that I had a lot of help with, so it would be great for us to document it with a big block comment.

nicbarker:
Courtesy of some great work from @FintasticMan https://github.com/nicbarker/clay/pull/119

![image](https://github.com/user-attachments/assets/54f33abd-bf6c-4834-b45d-b270aa5b801e)

nicbarker:
Explanation of the remaining (much smaller) `CLAY` macro has been added in this commit: https://github.com/nicbarker/clay/commit/a44423a1333290feb86b49d9f960fe8e41104ae1

VisenDev:
Nice!

PR #119, opened 2024-12-28
[Core] Simplify CLAY macro

FintasticMan:
While trying to debug the failing test in #118, I realised that this macro can be simplified quite a lot, and it can then also support more than 6 arguments!

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -90,26 +90,15 @@
 
 #define CLAY_STRING(string) (CLAY__INIT(Clay_String) { .length = CLAY__STRING_LENGTH(string), .chars = (string) })
 
-static int CLAY__ELEMENT_DEFINITION_LATCH = 0;
+static uint8_t CLAY__ELEMENT_DEFINITION_LATCH;
 
-#define CLAY__ELEMENT_INTERNAL(...) \
+// Publicly visible layout element macros -----------------------------------------------------
+#define CLAY(...) \
 	for (\
-		CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), __VA_ARGS__, Clay__ElementPostConfiguration(), 0); \
+		CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), ##__VA_ARGS__, Clay__ElementPostConfiguration(), 0); \
 		CLAY__ELEMENT_DEFINITION_LATCH < 1; \
 		++CLAY__ELEMENT_DEFINITION_LATCH, Clay__CloseElement() \
 	)
-
-#define CLAY__6_ARGS(a, b, c, d, e, f) a, b, c, d, e, f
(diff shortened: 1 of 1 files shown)


bullno1:
Isn't that a gcc extension?

clay.h
```diff
@@ -91,27 +91,14 @@
 
 #define CLAY_STRING(string) CLAY__INIT(Clay_String) { .length = CLAY__STRING_LENGTH(string), .chars = (string) }
 
-static int CLAY__ELEMENT_DEFINITION_LATCH = 0;
```
nicbarker, line 94: This looks great, only thing is I believe we need to keep this global definition here, because otherwise nesting the macro calls gives us a "declaration shadows local variable" warning for CLAY__ELEMENT_DEFINITION_LATCH 🤔

FintasticMan: It should be fine to shadow them, and I'm not even sure if it's possible to nest `CLAY` macros, because the nesting would have to take place in the for-loop initialisation. I don't think that you can put a for-loop there.

nicbarker: It actually produces an error in my local CLion with clang 🙂
The clay macros expand like this:

```C
CLAY() {
    CLAY() {

    }
}

// Expanded
for (
    uint8_t CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), Clay__ElementPostConfiguration(), 0);
    CLAY__ELEMENT_DEFINITION_LATCH < 1;
    ++CLAY__ELEMENT_DEFINITION_LATCH, Clay__CloseElement()
) {
    for (
    uint8_t CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), Clay__ElementPostConfiguration(), 0);
    CLAY__ELEMENT_DEFINITION_LATCH < 1;
    ++CLAY__ELEMENT_DEFINITION_LATCH, Clay__CloseElement()
    ) {
        // contents
    }
}
```

FintasticMan: Oh hmm, I don't any warnings for that at all, with clang or gcc with -Wall -Wextra -Wpedantic. What warning is it exactly?

nicbarker: Screenshot 2024-12-29 at 2 24 26 pm

FintasticMan: Ah ok, I'll put it back to being static then :-)

nicbarker: It seems to be under `-Wshadow` 🤔

FintasticMan: Hmm, even manually adding that flag doesn't seem to get it to show me that warning. Anyway, it does seem better to just keep it static if there's a chance it'll get someone's IDE to complain.

VisenDev:
this might address https://github.com/nicbarker/clay/issues/103

FintasticMan:
I've just rebased onto latest main, due to conflicts.

nicbarker:
Have tested this locally and it looks great to me, thanks for the hard work!

PR #120, opened 2024-12-28
[Examples/Intro] Fix NULL pointer deref due to huuge malloc

FintasticMan:
It would probably be good to check that malloc doesn't fail, but for a simple example like this it should be fine.


nicbarker:
Thanks so much for fixing this, was just testing something out and obviously left the test code in 🙈
Will be more careful next time!

PR #121, opened 2024-12-28
Move raylib renderer to single header

peterc-s:
As discussed on discord, to use Clay with the raylib renderer with GNU Make and GCC you had to separate out implementation from the `raylib_renderer_raylib.c` into a header file and make a few other changes.

I've moved the whole renderer into a single header file and this seems to work without any of those modifications needing to be made.

I've also fixed a segfault in the video example by copying over the minimum memory allocation from the raylib sidebar scrolling example.


peterc-s:
Sorry, realised my `git` was misconfigured and so commits came up as unverified so had to rebase and force push to sign them properly.

ThreedResident:
Any chance all this will be merged to main soon?

PR #122, opened 2024-12-28
[Renderers/Cairo] Add FindCairo.cmake

SuperOptimizer:
At least on my OSX system, even with `brew install cairo`, cairo isn't found by the C compiler in the standard locations. Calling `FindPackage` with this FindCairo.cmake script and then explicitly linking and including the library solves the issue.


nicbarker:
Fantastic work, I just ran this on my osx system and it works perfectly. Thank you!

PR #123, opened 2024-12-28
[Compilers] C projects should use C flags rather than CXX flags

SuperOptimizer:
In CMAKE, the `*CXX*` flags are for C++ only, so C files are not compiled with anything that falls under `CXX` flags. `C` flags should be used for `.c` files.


FintasticMan (commented):
I think it should use `CMAKE_C_FLAGS*` rather than `CMAKE_CC_FLAGS*`, see https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html.

SuperOptimizer:
> I think it should use `CMAKE_C_FLAGS*` rather than `CMAKE_CC_FLAGS*`, see https://cmake.org/cmake/help/latest/variable/CMAKE_LANG_FLAGS.html.

Huh, I could have sworn that Cmake used CCFLAGS. Maybe I'm thinking of normal make? Thanks for the tip, will fix

nicbarker:
Just waiting on the build then we can get this in 👍

nicbarker:
Looking good! 👍

PR #130, opened 2024-12-29
Add Clay_IsDebugModeEnabled()

juniorrantila:
This makes it easier to implement keyboard shortcuts that toggle the debug menu while also ensuring that the state does not get out of sync, like when closing the menu via the cross in the top right corner.


nicbarker:
Thank you @juniorrantila, this is a great call and I have needed to do it for a little while!

PR #132, opened 2024-12-29
Add GLFW3 + NanoVG example and renderer

juniorrantila:


nicbarker:
Just cloned this and had a look, it's looking fantastic already! I didn't know about nanoVG, love the super simple API!

juniorrantila:
Updated this to include the text measurement API changes from commits 326325f and e8025cc.

PR #134, opened 2024-12-30
Fixed compilation when using Clang on Windows

St0wy:
Clang on Windows defines `_MSC_VER` so using it for the definition of `CLAY_PACKED_ENUM` makes it use `__pragma` with Clang, which it cannot understand. This PR fixes it by replacing `#ifdef _MSC_VER` with `#if defined(_MSC_VER) && !defined(__clang__)`.


nicbarker:
Thank you for this, nice catch! I don't use Clang on windows so I haven't tested this myself, but we can always fix it again if it's not quite right 🙂

PR #135, opened 2024-12-30
[Core] Add "strict" mode for strict C99 compliance

FintasticMan:
Depends on #119.

This just allows users of the library to define `CLAY_STRICT` to get rid of the use of the non-standard `##__VA_ARGS__` construction, if they're OK with making sure that the affected macros aren't used without an argument.

I'm not sure if this is something we want, so I'd be very open to feedback! I like how much simpler the `CLAY_SIZING_*` macros are without having to worry about making sure that they also work when no arguments are given, so it might not even be out of the picture to just replace the current implementation with this strict one.


FintasticMan:
Superseded by #150.

PR #136, opened 2024-12-30
[Compilers] C++ projects should use CXX flags

SuperOptimizer:
It seems I made a whoopsie in #123 and edited the CXX flags of the example C++ project. This fixes that.


nicbarker:
Thanks, looks good!

PR #140, opened 2025-01-01
add RSGL renderer, RGFW backend and GLFW backend

ColleagueRiley:
RSGL is a single-header graphics library with a similar simple-to-use style as Raylib. 

RSGL itself does not handle rendering, it only batches out `RSGL_TRIANGLE` requests with draw data, making it easy to append low level APIs. 

It also doesn't care about the windowing backend, so it's pretty much a blank slate for any sort of rendering or windowing API. This PR includes a OpenGL implementation for RSGL as well as examples for both GLFW and RGFW. There is also currently a WIP software rendering backend for RSGL, but it wasn't included in this PR. 

The backend hasn't been fully implemented yet, and still has a lot of flaws. But I decided to create this initial PR for the backend to see if it interests anyone. 

NOTE: I include Makefiles instead of CMakefiles, those should probably be replaced as the project seems to only use CMake. Sadly, I'm not very familiar with CMake.


ColleagueRiley:
I added CMakeFiles, but they should be reviewed by someone more familiar with CMake.

PR #143, opened 2025-01-02
[Core] Fix default struct initialiser in C++

FintasticMan:
This means we don't need `-Wno-error=missing-braces` for the C++ project anymore.


nicbarker:
This is such a great idea for how to fix this problem! Awesome job 😁

PR #144, opened 2025-01-02
[Core] Add check for supported C/C++ versions

FintasticMan:


VisenDev:
Why restrict versions newer than C99? I don't think there have been any major breaking changes in the standard. 

(If anything, it seems like we should warn about using C89 accidentally, because that will definitely break and give some weird error messages)

FintasticMan:
This allows all C versions C99 or newer, there's a `!` not operator in front of the whole condition.

VisenDev:
Ahaha, I didn't see that, my bad

FintasticMan:
No problem, it's easily missed!

nicbarker:
Works like a charm, amazing work!

Issue #145, opened 2025-01-02
The clay.wasm included is not valid.


BrownNPC:
Running `wasm-validate` on the clay.wasm returns a `invalid magic number` error.

PR #146, opened 2025-01-02
build clay.wasm against master

BrownNPC:
The clay.wasm included in the repo is invalid.

this will fix https://github.com/nicbarker/clay/issues/145


nicbarker:
Hello, thanks for opening this. There are some changes in `clay.h` that haven't been carried over correctly into wasm just yet, so we might just wait until those are done 🙂
I will also need to make this change myself, just because I can't review binary files that are added to the repo, so it's necessary from a security perspective. Thanks for understanding!

PR #151, opened 2025-01-02
[Bindings/Odin] Update Odin bindings to latest

nicbarker:
This PR updates the Odin bindings to match the latest changes in `clay.h`, primarily adding the error handler to `Initialize`.


bindings/odin/clay-odin/clay.odin
```diff
@@ -64,11 +63,7 @@ ElementId :: struct {
     stringId: String,
 }
 
-when ODIN_OS == .Windows {
-    EnumBackingType :: u32
-} else {
-    EnumBackingType :: u8
-}
+EnumBackingType :: u8
```
nicbarker, line 66: We should now have working `u8` support on windows, but I will have to confirm this.

nicbarker: Just tested and I misunderstood, still needs to be setup this way

FintasticMan: I thought that the enums would be u8 on windows as well...

nicbarker: Yeah I was surprised and will investigate further, it's likely due to some combination of compiler and os that I'm using locally to generate the windows bindings. With the windows lib in the repo at the moment they are definitely 4 bytes, not 1 🤔

PR #152, opened 2025-01-02
[Core] Standardise number types to int32_t for array indices, lengths and capacities

nicbarker:
This PR modifies a number of internal and external number types that represent array indices, lengths and capacities to `int32_t`.

The reason that we've decided to use a signed type is that it's very common to do `<` and `>` comparisons with these fields, and using unsigned types creates a bit of a footgun where `0 - 1 > 0 == true`. It's very easy to accidentally decrement past zero, and this can create accidental logic bugs where comparisons return values that they shouldn't.


FintasticMan:
There are still a number of sign comparison issues on lines: 1410, 1513, 2992, 3255, 3257.

PR #153, opened 2025-01-02
[Core] Fix NULL pointer dereference when parent of floating container is invalid ID

peter15914:
The pointer ```parentItem``` could be NULL after the call to ```Clay__GetHashMapItem()```.
Therefore, I moved code blocks so that ```parentItem``` is checked for NULL before dereferencing.


nicbarker (approved):
Thank you, nice catch!

PR #155, opened 2025-01-03
[Core] Fix errors due to cast to same non-trivial type

FintasticMan:
Add a wrapper type with a single element so that both initialiser lists and variables can be passed to the config macros without requiring a cast.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -29,38 +29,36 @@
 
 // Public Macro API ------------------------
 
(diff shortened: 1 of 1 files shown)


nicbarker:
Thank you for this macro magic! I will play around with this locally and hopefully merge tomorrow 👍

nicbarker:
Looks good to me, thanks for the great work!

PR #162, opened 2025-01-04
[Renderers/Web] treat RenderCommand.commandType as uint8_t instead of uint32_t

bullyingteen:
This PR fixes the issue when RenderCommand.commandType has garbage if RenderCommand's padding memory is not zeroed.


nicbarker:
Thanks for the hard work tracking this down! 🙂
This LGTM as a short term fix, I definitely want to do another pass on how the types are handled with padding, as it feels pretty fragile at the moment.

PR #163, opened 2025-01-04
[Core] Bug in text wrapping at very narrow widths

### LINKED ISSUE #99: [Core] Bug in text wrapping at very narrow widths
Here I've used the debug tools to prevent the website demo from switching to the mobile layout properly, to demonstrate. For some reason, the words "flex-box" and "single" both repeat when it wraps, and they also "consume" the words beneath them! Obviously it's not exactly a huge problem here, but it may be indicative of a bigger issue so I thought it was worth it to raise.

image

bullyingteen:
fixes #99


nicbarker:
Absolutely amazing job finding this! 😁

PR #164, opened 2025-01-04
[Examples/Website] Fix CMakeLists.txt to compile to wasm

avinal:
## Changes

- CMake was not compiling to WebAssembly
- This changes adds missing CMake directives for compiling to Wasm


nicbarker:
Hello! Unfortunately this doesn't seem to work on all compilers / architectures, when I tried this originally on my machine I got:

`clang: error: unsupported option '-arch' for target 'wasm32'`

For some reason, CMake on my machine (OSX) always stuffs the `-arch` flag into the compiler flags, which clang doesn't like if compiling for the wasm32 target. I hope we can figure out a way of fixing it! 🤔

avinal:
Okay I do not have an OSX machine, let me investigate more.

PR #91, opened 2024-12-23
[Renderers/Terminal] Add initial implementation of terminal renderer

EmmanuelMess:
This is an initial approach to the terminal renderer. Its pretty bad, it does not take into account the "real" pixel size of the terminal, just the character size. 

Fixes #57

PR #168, opened 2025-01-06
Feature/terminal renderer

phyxl:
This builds on the work of @EmmanuelMess https://github.com/nicbarker/clay/pull/91/commits/7ce74ba46c01f32e4517032e9da76bf54ecf7a43

See the readme for more info, but this pull-request is for a terminal only renderer which utilizes ncurses which is found on nearly every linux distribution already


EmmanuelMess:
I would keep the ncurses renderer separate from the ANSI escape code based one.

renderers/ncurses/clay_renderer_ncurses.c
```diff
@@ -0,0 +1,129 @@
...
+            default: continue;
+        }
+    }
+    attr_on(color_set(0,0),0);
+    mvwprintw(win, 0, 0, "Number of color pairs used: %i", color_pair);
+    refresh();//update the screen/window
+}
+
+
+//written by EmmanuelMess: https://github.com/nicbarker/clay/pull/91/commits/7ce74ba46c01f32e4517032e9da76bf54ecf7a43
```
EmmanuelMess, line 106: Its probably not a good idea to add the author of the functions, git blame already provides this feature.

phyxl: It would, but I didn't use your branch, so it will show that I was the author instead. I want to make sure you get whatever credit you want for that.

nicbarker:
Hello, now that I've merged the basic ANSI terminal renderer example in https://github.com/nicbarker/clay/pull/91, I will get to this one next 🙂

PR #169, opened 2025-01-06
[Documentation] Updated example for Clay_SetPointerState

davidstyrbjorn:
Once again 😅🤓 found slightly out of date documentation in the `README.md` for examples when using `CLAY_IMAGE`
Seems like the struct was changed to use `Clay_Dimensions` so the examples no longer worked.


nicbarker:
Thank you @davidstyrbjorn, you're doing great work here 🤓

PR #174, opened 2025-01-08
[Core] Multi instance support

monodop:
* Moves global variables into Clay_Context
* Adds Clay_GetCurrentClayContext() and Clay_SetCurrentClayContext() to determine which context to use. This is stored in thread_local 
* Clay_Initialize automatically sets (and returns) the current context, so that you don't need to worry about it if you don't care about it
* All public and internal functions use GetCurrentClayContext() internally, to avoid polluting function signatures with new argument
* maxElementCount & max cache moved to arena, because I was having some issues getting the MinMemorySize function to work properly early on if it was in the context, but this can probably be fixed easily.
* maxElementCount & max cache can now only be set for initial values, they are made concrete within Clay_Initialize and cannot be changed for the arena/context after this point

This appears to be working on my local branch, but maybe could use some more testing

TODO:
- [x] update generators
- [ ] tests
- [x] add define statement in case thread_local doesn't exist?
- [x] click/hover callbacks probably need some more thought, could these ever be called with the wrong context?
- [x] I accidentally modified some signatures that shouldn't be modified. they should be reverted to the original signatures
- [x] update docs
- [x] move struct initializers to clay_initialize
- [x] Odin bindings updates


clay.h
```diff
@@ -1493,12 +1447,13 @@ Clay__DebugElementDataArray Clay__debugElementData;
...
 #endif
 
-Clay_LayoutElement* Clay__GetOpenLayoutElement(void) {
-    return Clay_LayoutElementArray_Get(&Clay__layoutElements, Clay__int32_tArray_Get(&Clay__openLayoutElementStack, Clay__openLayoutElementStack.length - 1));
+Clay_LayoutElement* Clay__GetOpenLayoutElement() {
+    Clay_Context* context = Clay_GetCurrentContext();
+    return Clay_LayoutElementArray_Get(&context->layoutElements, Clay__int32_tArray_Get(&context->openLayoutElementStack, context->openLayoutElementStack.length - 1));
 }
 
-uint32_t Clay__GetParentElementId(void) {
```
monodop, line 1500: Need to undo this change & use Clay_GetCurrentContext()

clay.h
```diff
@@ -1809,15 +1768,17 @@ Clay_LayoutElementHashMapItem *Clay__GetHashMapItem(uint32_t id) {
...
+    Clay_Context* context = Clay_GetCurrentContext();
+    Clay_LayoutElement *parentElement = Clay_LayoutElementArray_Get(&context->layoutElements, Clay__int32_tArray_Get(&context->openLayoutElementStack, context->openLayoutElementStack.length - 2));
     Clay_ElementId elementId = Clay__HashNumber(parentElement->childrenOrTextContent.children.length, parentElement->id);
     openLayoutElement->id = elementId.id;
     Clay__AddHashMapItem(elementId, openLayoutElement);
-    Clay__StringArray_Add(&Clay__layoutElementIdStrings, elementId.stringId);
+    Clay__StringArray_Add(&context->layoutElementIdStrings, elementId.stringId);
 }
 
-void Clay__ElementPostConfiguration(void) {
```
monodop, line 1819: Need to undo this change, & other similar ones where I accidentally removed void from signature

nicbarker:
👀

Issue #188, opened 2025-01-12
CLAY_TEXT not rendering lines that contain single characters


Ryzee119:
```
Clay_LayoutConfig layoutElement1 = {.padding = {4}, .sizing = {.height = CLAY_SIZING_FIT(), .width = CLAY_SIZING_FIT()}};
Clay_TextElementConfig textElement = {.fontId = FONT_ID_BODY_24, .textColor = {255, 255, 255, 255}, .fontSize = 64};
CLAY(CLAY_TEXT(CLAY_STRING("Hello\nWorld\nI\nCode\n"), CLAY_TEXT_CONFIG(textElement)), CLAY_LAYOUT(layoutElement1))
```
Seems somehow related to fontSize and how it scales?
Im using the raylib rendering backend

![image](https://github.com/user-attachments/assets/3280aabd-437a-4886-a220-95a63315fd34)

however changing to `fontSize = 32` (for example) appears to fix it

![image](https://github.com/user-attachments/assets/e4fa9830-d7bf-42d4-a3c3-019c362c8831)

nicbarker:
Hello 👋 
I think it's possible that you've got the syntax slightly confused, `CLAY_TEXT` is a child element and should go inside the braces, like this:

```C
Clay_TextElementConfig textElement = {.fontId = FONT_ID_BODY_24, .textColor = {255, 255, 255, 255}, .fontSize = 64};
CLAY(CLAY_LAYOUT(layoutElement1)) {
    CLAY_TEXT(CLAY_STRING("Hello\nWorld\nI\nCode\n"), CLAY_TEXT_CONFIG(textElement));
}
```

Ryzee119:
Thanks, still wrapping my head around the syntax, problems still occurs however

nicbarker:
@Ryzee119 that's totally fine! The syntax is a little odd at the moment for sure, we have plans to improve it 🙂

nicbarker:
Is it alright if I close this issue?

Ryzee119:
>Is it alright if I close this issue?

You can, but the issue is still present with the correct syntax too

Ryzee119:
https://github.com/nicbarker/clay/blob/208c7cb3a038307e8a7e26cf5057c16823d7d35a/clay.h#L1707

Changing this to `>= 1` fixes the issue.

nicbarker:
@Ryzee119 very nice work finding it!

nicbarker:
[...]

PR #190, opened 2025-01-13
Fixing rendering of lines that contain a single character

Ryzee119:
Fixes https://github.com/nicbarker/clay/issues/188

A line with a single character is length 1, but the check is only checking for lengths of 2 or more


nicbarker:
This one is superseded by https://github.com/nicbarker/clay/issues/188, I will attribute you as coauthor on the new PR 🙂

Issue #191, opened 2025-01-13
[Core] Strings that are shorter than the container size don't get parsed for new line characters


Ryzee119:
As a shortcut, clay will perform a quick check if the text will fit into the container unwrapped. This just pushes the whole string (unsliced) to the renderer.

https://github.com/nicbarker/clay/blob/208c7cb3a038307e8a7e26cf5057c16823d7d35a/clay.h#L2370-L2374

The issue with this is that is my string contains '\n' characters, these are not parsed and just get sent the the renderer. I dont believe this is the correct intent, this also results in the bounding box being miscalcuated.

Deleting this code block forces the text to be parsed correctly, however this would have a performance penalty too.

nicbarker:
Should be fixed when https://github.com/nicbarker/clay/pull/192 merges 🙂

nicbarker:
This should be fixed by https://github.com/nicbarker/clay/pull/192 🙂

PR #192, opened 2025-01-13
[Core] Fix text wrapping handling with explicit newline characters

nicbarker:
This is to address https://github.com/nicbarker/clay/issues/188. It fixes a bug where single character lines terminated by an explicit newline character were not correctly output, and also another bug that was hiding it: text wrapping handling was sometimes bypassed entirely in strings that contained newline characters.


nicbarker:
Supersedes https://github.com/nicbarker/clay/pull/190

Ryzee119:
Can confirm this fixes both https://github.com/nicbarker/clay/issues/191 and https://github.com/nicbarker/clay/issues/188

Thank you!

PR #200, opened 2025-01-14
[Core] Don't divide zero by zero

mikejsavage:
I was getting crashes here with floating point exceptions enabled.

This may not be a good fix but at least the resulting layout looks fine again, I can put together a repro if you think this is the wrong way to do it.


nicbarker:
Hey @mikejsavage, thanks for the good catch! I actually noticed another bug related to downsizing when I was investigating thing, would you be able to try out my change and confirm it fixes your bug correctly? 🙂

mikejsavage:
Yep your change works too!

nicbarker:
Awesome! Will merge this now.

PR #206, opened 2025-01-16
build clay-official-website with clang only

pvmm:
This commit also removes the obligatory clay subdirectory or allow the user to select another subdirectory.

Since clang is required for WebAssembly, stop clay-official-website from being compiled if GCC is selected. And now we force the WebAssembly build to be generated since the binary was not working anyway.

Files in **./examples/clay-official-website/build/clay** were removed since they are now copied automatically during build to **./build/examples/clay-official-website/web**.


nicbarker:
Hello, thanks for opening this PR 🙂
There are a couple of issues we're dealing with related to CMake & wasm, it's currently being tracked here: https://github.com/nicbarker/clay/pull/164. I'll close this one as a duplicate and you can track the other 👍

PR #207, opened 2025-01-16
Odin binding generator script

monodop:
I generalized my clay.h parser and wrote a generator for odin bindings. Ideally we could write different generators for different output bindings to reduce the amount of work needed to keep everything up to date. For odin, there's a pretty large number of weird exceptions and edge cases, so it might be worthwhile standardizing them.

This generator/parser might be more effort than it saves though, curious to hear thoughts about it.


monodop:
Also for reference, here's what the extracted symbols looks like (from parser.py) https://gist.github.com/monodop/6875a1d677eb85212494e21025f4bff0

PR #212, opened 2025-01-19
SetMesureText and SetQueryScrollOffset takes userData

### LINKED ISSUE #63: [Core] UserData for Clay_SetMeasureTextFunction
So this function

```C
void Clay_SetMeasureTextFunction(Clay_Dimensions (*measureTextFunction)(Clay_String *text, Clay_TextElementConfig *config));
```

There is no `userData` being passed to compared to the callback to

```C
void Clay_OnHover(void (*onHoverFunction)(Clay_ElementId elementId, Clay_PointerData pointerData, intptr_t userData), intptr_t userData);
```

I wonder if it would make sense to add that?

emoon:
Fixes #63 

This is a breaking change due to new parameter. I set it to `0` in all cases except the `SDL2` case where I used it as an example of passing the font data.

Another change now is because the userData is set inside the context object `Clay_Initialize` has to be called before calling this.


nicbarker:
Thank you! I was literally implementing this last night 😁
I'll just implement passing the raylib fonts through in the same way as the SDL ones and we should be good to go.

> Another change now is because the userData is set inside the context object Clay_Initialize has to be called before calling this.

Yeah, it's not super obvious the required order. Soon I think we'll change initialize to take a config struct that includes both the dimensions and the measure text function, which will get auto init to NULL if not provided 😁

nicbarker:
Actually, I think it's fine to just merge this now and fix up the raylib example later - I'd like to switch the fontId to a `uintptr_t` anyway, as it's currently not very ergonomic to use.

Issue #213, opened 2025-01-19
[CMake] Don't pull massive 3rd party dependencies via CMAKE by default


Robert-M-Muench:
See: https://discord.com/channels/1291501630973415474/1328824444009054291/1330582239721029735

CLAY is announced as non-dependency lib, but pulls a ton of 3rd party dependencies via CMAKE, which pollutes/conflicts other projects using CLAY.

Make all those dependencies, like SDL3, optional or guard them.

Robert-M-Muench:
The main culprit on my side was that I used *SDL3* as the CMAKE target name, and CLAY uses *SDL*.

It would make sense to list the target names in the build section. And maybe be more specific if it's SDL3 or SDL2.

nicbarker:
Hello, thanks for opening this issue 👋
Could you give me a little more detail on what you're trying to achieve here?
Clay doesn't have a "build" per se, given that the library itself is a single file with no dependencies.
The `CMakeLists.txt` in the base of the repo is just for building the examples, but in your own project I think it just makes sense to pull in clay.h and not much else 🙂

Robert-M-Muench:
Yes, I understand.

The thing is when you are working on a bigger project using CMAKE. What you usually do is something like this:

```cmake
FetchContent_Declare(
        clay
        GIT_REPOSITORY https://github.com/nicbarker/clay.git
        GIT_TAG main
)
FetchContent_MakeAvailable(clay)
```
This makes it super simple to have CLAY available, the IDE finding all necessary files, etc.

In bigger projects it's more convenient to use CMAKE for all the dependencies that adding a single .h file manually. Further, it simplifies using specific release-/tag-versions.

Even CLAY being a single-header-only thing, I think it makes sense to provide a CMAKE compatible way of including it via *FetchContent_Declare*.

pope:
Is the answer to create an OPTION that only builds examples if the option is set to true?

```cmake
option(CLAY_BUILD_EXAMPLES "Build example programs" OFF)

# ...

if(CLAY_BUILD_EXAMPLES)
  add_subdirectory("examples/cpp-project-example")
  # ...
endif()
```

nicbarker:
This has been implemented in #216, thanks for the work everyone!

PR #216, opened 2025-01-20
[CMake] Make Examples Optional in CMAKE

CJFEdu:
Makes including the examples optional.  SDL2/SDL3 specifically have long download/build times.  If they aren't being used, then this can dramatically speed up build times.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,21 +3,38 @@ project(clay)
 
 list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
 
-option(CLAY_INCLUDE_EXAMPLES "Build examples" OFF)
+option(CLAY_INCLUDE_ALL_EXAMPLES "Build all examples" OFF)
+option(CLAY_INCLUDE_DEMOS "Build video demo and website" ON)
+option(CLAY_INCLUDE_CPP_EXAMPLE "Build C++ example" OFF)
+option(CLAY_INCLUDE_RAYLIB_EXAMPLES "Build raylib examples" OFF)
+option(CLAY_INCLUDE_SDL2_EXAMPLES "Build SDL 2 examples" OFF)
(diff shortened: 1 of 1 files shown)


FintasticMan:
I'm not sure I understand what use case exists for running CMake with the flag disabled... The only thing the CMake project does is compile the examples, so with it disabled, CMake does nothing at all.

CJFEdu:
I'm not a CMAKE expert by any means, so I could be doing the following wrong.  I'm using FetchContent to add Clay to my project.  This adds the examples, which then adds all the dependencies. 

That being said, this was just a quick solution to my specific problem.  A more practical solution would probably be to have the option take a string to allow you to select which example you want to build.  For example, someone that only wants to build the Raylib example shouldn't be required to download and build both versions of SDL.

Robert-M-Muench:
See #213 same problem, same suggestion.

nicbarker:
Thanks for figuring this out! I'm fine with this as long as we flip the default option to `ON`, as I think it could be confusing if someone clones the repo and can't figure out why the examples won't build by default.

CJFEdu:
I added an update that makes the CMAKE build options more granular.  I set build all examples to OFF as a default since downloading all of SDL 2 and 3 seems excessive.  Instead, the demo from the video and the website are turned on by default.

These options should probably be included in the ReadMe somewhere, but I wasn't sure where the logical place to put it would be.

nicbarker:
Sorry about the back and forth, but I think it would be best if we left the examples `ON` by default with the option to switch them off, both to prevent confusion and also because we build the examples through github actions on PRs to make sure they work 🙂

Robert-M-Muench:
That's OK as well. As long as there is a way I can use the CMAKE stuff and configure it somehow to only use the CLAY code.

CJFEdu:
@nicbarker  Done.

nicbarker:
Looks good to me, thanks for the work here! 🙂

PR #232, opened 2025-01-28
[Renderers/SDL2] Don't take addresses of temporaries.

radiant64:
Fixes compilation errors when including the renderer in C++ sources.


FintasticMan (commented):
The extra block isn't needed, as it doesn't affect the lifetime of the `rect` variable.

radiant64:
> The extra block isn't needed, as it doesn't affect the lifetime of the `rect` variable.

You're absolutely right about that; removed.

PR #234, opened 2025-01-28
Clay__defaultMaxMeasureTextWordCacheCount is not used

Christian-Schl:
The Clay__defaultMaxMeasureTextWordCacheCount is not use in the function Clay_Initialize. By now for the WordCacheCount value was element count * 2. This Mr fix it and use the correct variable for init the context. 

Without the changes Clay_Initialize can fail for non default values calculated by Clay_MinMemorySize

Thx


nicbarker:
Thanks for reporting this! It appears that great minds think alike, and https://github.com/nicbarker/clay/pull/233 was just merged that makes the same change 🙂

PR #236, opened 2025-01-29
Define CLAY_IMPLEMENTATION in Jetbrains IDE

monodop:
[`__JETBRAINS_IDE__`](https://blog.jetbrains.com/clion/2017/02/clion-2017-1-eap-debugger-fixes-ide-macros-and-new-cmake/) is defined within Jetbrains editors like CLion and Rider. This is only defined inside the IDE, and does not affect builds, so we can use it to enable `CLAY_IMPLEMENTATION` to improve debugging & intellisense in this environment. I'm not sure if other IDEs have similar macros, but I figured we could start with this one.

Before:
image

After:
image

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -21,6 +21,11 @@
 #include <arm_neon.h>
 #endif
 
+#ifdef __JETBRAINS_IDE__
+// Help jetbrains IDEs like CLion and Rider with intellisense & debugging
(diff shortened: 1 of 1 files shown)


nicbarker:
Apologies this took so long to get merged, just tested it out and it looks great!

PR #237, opened 2025-01-29
Added C3 bindings for Clay

JeffereyAEL:
That's it. Thought it would be cool


KaruroChori:
Even if they don't get merged, thanks :D. I have a new excuse to use c3 :).

nicbarker:
This is amazing work @JeffereyAEL! Thanks for putting in the effort here, will try to get this merged early next week 👍

PR #239, opened 2025-01-29
Copy elementId in Clay__AddHashMapItem() in case underlying stringId has changed

monodop:
If the string reference passed to an element id has changed between frames (e.g. moved), then the reference stored in the hash item can become stale. This change makes sure that we're always using the latest value.


nicbarker:
Great catch! Thanks 🎉

PR #240, opened 2025-01-30
[Core] Replace config macros with a single unified configuration struct

nicbarker:
This PR replaces the element configuration macros like `CLAY_RECTANGLE` with a single unified configuration struct:

```C
typedef struct {
    Clay_ElementId id;
    Clay_LayoutConfig layout;
    Clay_RectangleElementConfig rectangle;
    Clay_ImageElementConfig image;
    Clay_FloatingElementConfig floating;
    Clay_CustomElementConfig custom;
    Clay_ScrollElementConfig scroll;
    Clay_BorderElementConfig border;
} Clay_ElementDeclaration;
```


FintasticMan (commented):
I really like this idea! It should also make bindings to other languages easier.

clay.h
```diff
@@ -126,7 +118,7 @@ static uint8_t CLAY__ELEMENT_DEFINITION_LATCH;
 */
 #define CLAY(...) \
 	for (\
-		CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), __VA_ARGS__, Clay__ElementPostConfiguration(), 0); \
+		CLAY__ELEMENT_DEFINITION_LATCH = (Clay__OpenElement(), Clay__ConfigureOpenElement(CLAY__INIT(Clay_ElementDeclaration) __VA_ARGS__), 0); \
```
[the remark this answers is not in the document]
FintasticMan: And maybe `OpenElement` could take the config struct as parameter, rather than a separate function?

emoon:
Shouldn't https://github.com/nicbarker/clay/blob/main/bindings/odin/clay-odin/clay.odin also be updated the API has changed?

clay.h
```diff
@@ -1436,7 +1395,55 @@ void Clay__CloseElement(void) {
...
+            uint64_t mask = vgetq_lane_u64(vreinterpretq_u64_u8(cmp), 0) &
+                            vgetq_lane_u64(vreinterpretq_u64_u8(cmp), 1);
+
+            if (mask != UINT64_MAX) { // If there's a difference
+                for (int32_t i = 0; i < 16; i++) {
+                    if (s1[i] != s2[i]) {
+                        return false;
+                    }
+                }
+            }
```
emoon, line 1426: I think this can be simplified down to (untested)
```C
if (vminvq_u32(mask) != 0xFFFFFFFF) { // If any element is not equal
    return false;
}
```

emoon: Another approach is to use 
```
const uint8x8_t res = vshrn_n_u16(mask, 4);
const uint64_t matches = vget_lane_u64(vreinterpret_u64_u8(res), 0);
```
this gives you a `u64` value where each nibble is 4 bit representation of the compare mask. ref https://community.arm.com/arm-community-blogs/b/servers-and-cloud-computing-blog/posts/porting-x86-vector-bitmask-optimizations-to-arm-neon

nicbarker: Awesome, thanks for the tips!

clay.h
```diff
@@ -1551,9 +1668,6 @@ void Clay__CompressChildrenAlongAxis(bool xAxis, float totalSizeToDistribute, Cl
         float targetSize = 0;
         for (int32_t i = 0; i < resizableContainerBuffer.length; ++i) {
             Clay_LayoutElement *childElement = Clay_LayoutElementArray_Get(&context->layoutElements, Clay__int32_tArray_GetValue(&resizableContainerBuffer, i));
-            if (!xAxis && Clay__ElementHasConfig(childElement, CLAY__ELEMENT_CONFIG_TYPE_IMAGE)) {
-                continue;
-            }
```
nicbarker, line 1556: This is now handled by not even adding these containers into the resizeableContainerBuffer in the first place

Issue #231, opened 2025-01-28
clay_interactions.h


monodop:
There's been a bunch of discussions in the discord server, and in the github issues (#230, #224, #198, #197, #154, #67) about moving mouse/scroll interactions out of clay.h and leave it for the user space. While I do personally think that clay should come out of the box with support for mouse/keyboard/scroll inputs (as well as accessibility features like tab indexes, keyboard navigation, etc), I wanted to start a thread about how a theoretical input system could be developed on top of clay.h. I'll be referring to a theoretical header called `clay_interactions.h`, which could be bundled in the clay repo. But even if not, I think this is still a useful exercise, as anyone trying to implement this will still have similar requirements.

> Note: to be clear, this is mostly a brain dump containing most of my thoughts on this over the last couple weeks, and I'm really interested to hear others' opinions.

## Things that I'm intentionally leaving out of scope here:
* Text inputs / text layout queries / etc. While this is definitely related to user io functionality, I think we should probably have a solution for rich text and/or masonry layout first.

## State Management
One of the first problems that needs to be solved, is how could such a library store its state? For example:

* For clicks, which elements has the click started in
* For hover, which elements is the mouse entering / leaving
* How long has it been since a click? (e.g. double click)

The library can of course have its own memory storage, like having its own arena. However, this means that the consumer would need to allocate multiple blocks of memory and track them separately, and take extra caution not to mix & match contexts with different interaction memory.

### Proposal for clay.h

* Allow the ability to somehow extend `Clay_Context` to include additional third party data, preferably in a generic/extensible way. This would allow third party libraries to keep their data co-located with the clay context, and could be re-allocated at the same time as the rest of the memory, e.g. if max element count needs to be increased. 
* (maybe?) Allow the ability to extend the element hash map entry (different from custom element configs), allowing libraries to store additional element state and have it get cleaned up automatically.

## Querying Layout Information
There's a lot of tasks that an interactions library might need to perform based on the final layout data from the previous frame:

* Get an ordered list of elements at an arbitrary point or set of points. Example: click/hover; drag & drop; local multiplayer game with multiple cursors.
* Find elements that are adjacent to a given element. Example: use arrow keys to navigate an inventory; find the next text element to continue text selection into
* Get the bounding box of an element. Example: determine height of scroll bar for a partially obscured element; get target information for an animation
[...]

PR #241, opened 2025-01-30
Clay_GetElementIdsAtPoint

monodop:
regarding #231 - `Get an ordered list of elements at an arbitrary point or set of points. Example: click/hover; drag & drop; local multiplayer game with multiple cursors.`

I wasn't sure what we wanted to do with SetPointerState - I wanted a version of this function that I could use arbitrarily to do the same behavior, except I don't want it to do any hover callbacks or store any information about pointer state.

As a result, I made this version, which is basically just copy/pasted but with those parts removed. Maybe this could be refactored so that SetPointerState calls this?

In order to get this working, I needed a place to store the element list in the first place. The approach I went with here is to just create a pointQueryIds list, where results will be stored. The downside is that if you call this multiple times, you'll need to manually save the results yourself outside of clay. I'm open to suggestions on a better approach, but this does seem to work.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -510,6 +510,12 @@ typedef struct {
     uintptr_t userData;
 } Clay_ErrorHandler;
 
+CLAY__TYPEDEF(Clay_PointQueryResult, struct
+{
+    int32_t length;
+    const Clay_ElementId *results;
+});
+
 // Function Forward Declarations ---------------------------------
 // Public API functions ---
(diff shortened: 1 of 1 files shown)


monodop:
In order to avoid confusion, maybe we should add a second method to retrieve the pointerOverIds from SetPointerState? I'm still not sure how I feel about having these two separate functions that work independently.

PR #245, opened 2025-01-31
[Renderers/SDL2] Add rounded rectangle support to sdl2 renderer; feature-completes sdl2 renderer

steviegt6:
Ports over the existing rounded rectangle-rendering code for the SDL3 renderer to the SDL2 renderer, completing its missing features list.


nicbarker:
Fantastic work, thank you so much for this!

PR #208, opened 2025-01-17
Extend SDL2 Renderer and SDL2-video-demo

ppebb:
This PR Implements CLAY_RENDER_COMMAND_TYPE_IMAGE and CLAY_RENDER_COMMAND_TYPE_BORDER for SDL2. It uses SDL_image for loading an example PNG.

For CLAY_IMAGE, the imageData provided is just the SDL surface returned by SDL_image's load functions.

Corner radii have not been implemented because drawing circular geometry with SDL2 requires a lot of work from any examples I found. Perhaps it can come in a later PR.

Both elements have been added to SDL-video-demo to provide examples, although I wasn't sure where to put them or what to do so I just put them in the top left corner.

PR #246, opened 2025-01-31
[Renderers/SDL3] Add image rendering and scissor support to SDL3 renderer

steviegt6:
Implements `CLAY_RENDER_COMMAND_TYPE_IMAGE` for the SDL3 renderer using SDL_image essentially identically to how it's done in SDL2 (see #208).

Also implements `CLAY_RENDER_COMMAND_TYPE_SCISSOR_START`/`CLAY_RENDER_COMMAND_TYPE_SCISSOR_END` based on the similar implementation in the SDL2 renderer.


steviegt6:
When I first implemented images, the example was split and I just inserted an image to render. Now the SDL3 example uses the video demo and the image example in the SDL2 example also goes unused, I'll leave the loading there as an example but I'm not sure what the ideal change to make it.

renderers/SDL3/clay_renderer_SDL3.c
```diff
@@ -232,6 +235,29 @@ static void SDL_RenderClayCommands(SDL_Renderer *renderer, Clay_RenderCommandArr
...
+            case CLAY_RENDER_COMMAND_TYPE_SCISSOR_END: {
+                SDL_SetRenderClipRect(renderer, NULL);
+                break;
+            }
+            case CLAY_RENDER_COMMAND_TYPE_IMAGE: {
+                Clay_ImageRenderData *config = &rcmd->renderData.image;
+                const SDL_Texture *texture = SDL_CreateTextureFromSurface(renderer, config->imageData);
+                const SDL_FRect dest = { rect.x, rect.y, rect.w, rect.h };
+
+                SDL_RenderTexture(renderer, texture, NULL, &dest);
```
ernestoyaquello, line 258: After it's rendered, we should destroy the texture here:

```suggestion
                SDL_RenderTexture(renderer, texture, NULL, &dest);
                SDL_DestroyTexture(texture);
```

ernestoyaquello:
Is there a reason why we pass a surface and then convert it to a texture? The SDL_image library allows us to get a texture directly for any image (`IMG_LoadTexture`, etc), so that seems much more efficient. 🤔

nicbarker:
Thank you for the great work here, we just get the DestroyTexture in to prevent the memory getting leaked, then we can merge it 😁

steviegt6:
Thanks for merging; sorry I didn't get to the additional changes myself! I've been busy with work.

nicbarker:
That's totally fine, thank you for the good work!

PR #247, opened 2025-02-01
[Bindings/csharp] Create csharp bindings README

Orcolom:


nicbarker:
Hello, and great work here!

PR #251, opened 2025-02-06
[Core] add CLAY_DISABLE_SIMD flag to conditionally disable SIMD includes

johan0A:
`#include <emmintrin.h>` was problematic for my use case as emmintrin.h depends on libc. 
I propose adding a CLAY_DISABLE_SIMD flag to conditionally disable SIMD includes to solve this kind of issues.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -15,9 +15,9 @@
 #include <stddef.h>
 
 // SIMD includes on supported platforms
-#if defined(__x86_64__) || defined(_M_X64) || defined(_M_AMD64)
+#if !defined(CLAY_DISABLE_SIMD) && (defined(__x86_64__) || defined(_M_X64) || defined(_M_AMD64))
 #include <emmintrin.h>
-#elif __aarch64__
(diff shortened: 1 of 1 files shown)


nicbarker:
Thanks, this is a good idea and I should have included it initially. Just OOI, which compiler are you using? If you're generating a lib for zig to use via clang, you should be able to avoid the libc include by using `-ffreestanding` flag

emoon:
I ran into this issue as well. Because the Rust bind generator does recursive visit of the included files when generating bindings and `emmintrin.h` included stdlib.h and other stuff which drag into pthread symbols and what not :( I managed to get rid of most of it as the bindgen tool allow you to blacklist with a regex, but a define will help more.

johan0A:
> you should be able to avoid the libc include by using `-ffreestanding` flag
nice, ill do that instead then, thanks

nicbarker:
It's something we'll need to make clear to library maintainers, because it's not an obvious or well understood flag 😅

PR #252, opened 2025-02-07
[Bindings/Odin] Add get/set current context method to Odin bindings

nadako:
This is useful for hot-reloading and was missing from the bindings. I don't know how much of the Context you want to expose, but I assume it should be opaque.


nicbarker:
Hello, thanks for the work! We indeed haven't settled on a proper public / private API distinction in context, so will be best to leave opaque for now 🙂

PR #253, opened 2025-02-07
Odin Bindings: simplify clay.UI usage, just pass ElementDeclaration directly instead of requiring an extra .configure call

nadako:
`configure` was introduced in 9d940c1f8e4a3d38d6f3ecdf51da2990f8b2f5e2, but it feels a bit pointless and boilerplatey to write `UI().configure({...})` when we could just do `UI({...})`.

Correct me if I'm wrong and there are actually use-cases for `ClayOpenElement`

### DIFF
--- a/bindings/odin/clay-odin/clay.odin
+++ b/bindings/odin/clay-odin/clay.odin
@@ -368,19 +368,11 @@ foreign Clay {
     _GetOpenLayoutElementId :: proc() -> u32 ---
 }
 
-ClayOpenElement :: struct {
-    configure: proc (config: ElementDeclaration) -> bool
-}
-
-ConfigureOpenElement :: proc(config: ElementDeclaration) -> bool {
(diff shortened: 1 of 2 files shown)


farism:
I believe this was changed to support `Clay_Hovered` and friends inside of the ElementDeclaration. This ensures the element is open before trying to call those stateful helpers.

nicbarker:
Hello, yes @farism is correct here - we now properly support the `Clay_Hovered()` function from the C API in odin, which allows you to test if an element is hovered as part of it's declaration: https://github.com/nicbarker/clay/commit/dcd6feda86cf09e7b26c980774661dbd0b9b5481

nadako:
I see, that makess sense, but is also a bit unfortunate that we have to write out these configure calls. Might be a matter of habit, but it kinda breaks the "almost-declarative" structure for layouts for me.

Issue #257, opened 2025-02-13
[Renderer/SDL2] Add rounded corner borders and remove duplicate right border. 


TimothyHoytBSME:
I'm currently trying my hand at this, but having issues. Mimicking the DrawRing function from RayLib, but it is a bit jagged and doesn't quite cover the rounded rectangle's corners.

nicbarker:
I believe this has been fixed by #258 😁

PR #258, opened 2025-02-13
[Renderers/SDL2] add rounded corner borders and fixed maximum radius issues.

TimothyHoytBSME:
Added a function to create rounded corner borders, in the same fashion as the rounded corner rectangle function.
Changed the border render section to call the new function for the four corners and to account for bounding box being smaller than combined border widths.
Also fixed one typo and added a conditional M_PI define if cmath constant is not present.

Fixes issue [257](https://github.com/nicbarker/clay/issues/257).


CrossScarDev:
I hope this gets merged soon

nicbarker:
Thanks for the great work here!

PR #259, opened 2025-02-14
Support older c++ versions

Amronos:
I updated the ``#if`` statement that checks for supported C/C++ versions to support older c++ versions. Which I do think Clay supports.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -29,11 +29,11 @@
 #define CLAY_HEADER
 
 #if !( \
-    (defined(__cplusplus) && __cplusplus >= 202002L) || \
+    (defined(__cplusplus) && __cplusplus >= 201103L) || \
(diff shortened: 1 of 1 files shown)


FintasticMan:
Unfortunately, designated initialisers (3 in this list https://en.cppreference.com/w/cpp/language/aggregate_initialization) is only supported from C++20 onwards. We make heavy use of these and I'm quite sure we're not willing to get rid of their use.

nicbarker:
> Unfortunately, designated initialisers (3 in this list https://en.cppreference.com/w/cpp/language/aggregate_initialization) is only supported from C++20 onwards. We make heavy use of these and I'm quite sure we're not willing to get rid of their use.

This is correct, it's a possibility down the track once the core library stabilises enough, but at the moment it would be too much of a disruption to the development workflow 🥲

PR #261, opened 2025-02-14
[Renderers/SDL3] Enable sdl3 alpha blending

irfan-zahir:
SDL_SetRenderDrawBlendMode(renderer, SDL_BLENDMODE_BLEND) enables alpha blending in SDL's renderer. Without this, alpha values in colors won't work properly - everything will be rendered as fully opaque regardless of the alpha value.


nicbarker:
Thank you for this fix!

PR #266, opened 2025-02-16
[Renderers/Raylib] Convert Image usage to Texture

CrackedPixel:
Some people in Discord were getting confused about this:

I dont think `Image` is meant to be used here but instead `Texture` (as used correctly for the Clay website, etc)
`Image` is on the CPU, `Texture` is on the GPU

Additionally,
`LoadImage()` + `LoadTextureFromImage()` without `UnloadImage()` causes a memory leak
raylib has a separate `LoadTexture()` that does the load+send to gpu+unload for you to simplify this

### DIFF
--- a/README.md
+++ b/README.md
@@ -1240,7 +1240,7 @@ Used to perform **aspect ratio scaling** on the image element. As of this versio
 
 ```C
 // Load an image somewhere in your code
-Image profilePicture = LoadImage("profilePicture.png");
+Texture profilePicture = LoadTexture("profilePicture.png");
 // Note that when rendering, .imageData will be void* type.
 CLAY({ .image = { .imageData = &profilePicture, .sourceDimensions = { 60, 60 } } }) {}
 ```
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -1249,15 +1249,15 @@ CLAY({ .image = { .imageData = &profilePicture, .sourceDimensions = { 60, 60 } }
...
+Texture profilePicture = LoadTexture("profilePicture.png");
 // Declare a reusable image config
 Clay_ImageElementConfig imageConfig = (Clay_ImageElementConfig) { .imageData = &profilePicture, .sourceDimensions = {60, 60} };
 // Declare an image element using a reusable config
 CLAY({ .image = imageConfig }) {}
 // Declare an image element using an inline config
 CLAY({ .image = { .imageData = &profilePicture, .sourceDimensions = {60, 60} } }) {}
 // Rendering example
-Image *imageToRender = renderCommand->elementConfig.imageElementConfig->imageData;
+Texture *imageToRender = renderCommand->elementConfig.imageElementConfig->imageData;
```
CrackedPixel, line 1260: `imageData` is a `void*` so this should still work? @nicbarker

TimothyHoytBSME:
The only issue I see, is the readme covers other renders, (not raylib specifically) that do not use Textures.

CrackedPixel:
> The only issue I see, is the readme covers other renders, (not raylib specifically) that do not use Textures.

Yep, that is a good point. What do you think of it now? I figured it was meant to be whatever format other renders/loaders use and so it would make it more clear they can use what they want

TimothyHoytBSME:
It's a tough call here, one Nic will likely need to make. Personally, I would probably leave Image in the readme but have a comment and/or description about Texture for RayLib anytime image is mentioned... I'm really not sure what would be best for the general population, or for the future.

I would leave it like this and let Nic have a look.

README.md
nicbarker: Yep this looks all good 👍

nicbarker:
Just tried it out and it works great, thanks for the work here!

PR #267, opened 2025-02-16
[Renderers/Allegro4] Add Allegro 4.x renderer and demo

eigenlenk:
# Allegro 4 renderer

#### API conformance
- [x] Rectangle (+rounded)
- [x] Text
- [x] Border (+rounded)
- [x] Scissor
- [x] Custom (accepts a function pointer to allow custom drawing within element)

#### Missing or incomplete features
- [ ] Image (missing resizing support and corner rounding)

# Allegro 4 demo

Demo is rather random/silly and based on this screenshot of Microsoft Encarta 95:
image

Remade in Clay:
image

Same demo compiled for MS-DOS:
image

---

So yeah, perhaps not the most useful renderer and demo since 4.x is rather old. But now that I have a better grasp of the library, I could try supporting the more modern 5.x version which is still actively developed. I also couldn't work out fetching and compiling the library through Make, so a pre-built binary is included instead 🤷‍♂️


nicbarker:
Hello, sorry for the delay in getting to this. It looks so awesome 😁
Don't worry about not including the .dll, I'm sure people who are motivated to use the Allegro renderer will be able to source a copy of the lib themselves 🙂
Thank you for the hard work!

eigenlenk:
I think I need to drop Make support for this and include a README instead. It's not going to build on Linux out of the box anyway, and only works with MinGW on Windows as well. Should silence these failing checks.

PR #269, opened 2025-02-17
Make Clay_MinMemorySize accurate based on word cache count

alecks:
This PR just ensures that MinMemorySize is accurate according to the `maxMeasureTextWordCacheCount` by ensuring `fakeContext` shares the same value as the `currentContext`. Found this after constantly running out of memory when dynamically adding lots of text.

Thanks for the amazing library!


nicbarker:
Hello, great catch! Thanks for picking this up 🙂

dazKind:
I cloned 3 hours ago and was about to PR this. You guys rock! :D

PR #175, opened 2025-01-08
[Bindings/Odin] Added missing public and private API function bindings

cstrachan88:
Added the missing bindings to the Odin source. All of the added procedures work as expected. With this, I was able to build an interactive example following the youtube tutorial. There is one issue getting this to work. Using `Hovered` as a condition to set the rectangle color unfortunately tests against the parent element, so all of the document title rectangles are highlighted. This can be seen at the start of the video. Everything else matches the youtube example, include the file menu.

https://github.com/user-attachments/assets/16d3fc40-e7d2-4b96-9e17-80bdd2680920

```odin
for d, i in documents {
  if u32(i) == selected_document_idx {
    if clay.UI(
      clay.Layout({padding = {16, 16}, sizing = clay.Sizing{width = clay.SizingGrow({})}}),
      clay.Rectangle({color = {120, 120, 120, 255}, cornerRadius = {8, 8, 8, 8}}),
    ) {
      clay.Text(d.title, clay.TextConfig({fontId = DEFAULT_FONT_ID, fontSize = 16, textColor = COLOR_WHITE}))
    }
  }
    else {
    if clay.UI(
      clay.Layout({padding = {16, 16}, sizing = clay.Sizing{width = clay.SizingGrow({})}}),
      // BUG: Hovered is triggering when the parent container is hovered
      clay.Rectangle({color = clay.Hovered() ? {120, 120, 120, 120} : {}, cornerRadius = {8, 8, 8, 8}}),
    ) {
      // NOTE: OnHover must be handled inside element - doesn't return a TypedConfig
      clay.OnHover(handle_hover, nil)
      clay.Text(d.title, clay.TextConfig({fontId = DEFAULT_FONT_ID, fontSize = 16, textColor = COLOR_WHITE}))
    }
  }
}
```

and

```odin
handle_hover :: proc "c" (elementId: clay.ElementId, pointerInfo: clay.PointerData, userData: rawptr) {
  context = state.ctx
  // NOTE: Using offset here since it is sequential from 0 and matches the document index
  if pointerInfo.state == .PRESSED_THIS_FRAME {
    selected_document_idx = elementId.offset
  }
}
```

PR #270, opened 2025-02-17
[Bindings/Odin] add missing bindings, fix binding, improve ergonomics of userdata, conform to stricter style flags

laytan:
- Adds bindings for all public APIs in clay.h to the Odin bindings.
- Fixes the binding for `Initialize`, it should return a pointer to the
  context.
- Use `rawptr` for userdata arguments, it's a more ergonomic type for
  this use case.
- Many Odin users build with these flags (-vet -strict-style) and it would turn up errors here. 
   It is therefore advised to build libraries to conform to the strictest Odin flags.

# Additional improvements

- Odin prefers tabs for indentation, but most of the bindings currently do spaces, any reason for this? There is even an Odin flag to strictly check for tabs with `-vet-tabs`
- `clay.UI(}.configure({ id = clay.ID("foo") })` could be made to be `clay.UI()({ id = clay.ID("foo") })`, by just returning the function directly instead of wrapped in a struct, I don't mind the current way that much, but thought I should mention it.

### DIFF
--- a/bindings/odin/clay-odin/clay.odin
+++ b/bindings/odin/clay-odin/clay.odin
@@ -3,461 +3,457 @@ package clay
 import "core:c"
 
 when ODIN_OS == .Windows {
-    foreign import Clay "windows/clay.lib"
+	foreign import Clay "windows/clay.lib"
 } else when ODIN_OS == .Linux {
-    foreign import Clay "linux/clay.a"
+	foreign import Clay "linux/clay.a"
 } else when ODIN_OS == .Darwin {
-    when ODIN_ARCH == .arm64 {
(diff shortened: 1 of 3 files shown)


laytan:
Ah I just saw #175 , that's embarrassing. I think this is more ready to go, but I don't want to take away credit/work from there, so if you want to continue that PR just close this.

nicbarker:
All good, we can co-author it and the API has drifted a lot since that PR was opened 👍

nicbarker:
This is a fantastic idea to reduce the boilerplate btw! Will be great not having `configure` repeated everywhere!

Issue #271, opened 2025-02-17
[Core] Can't set userData on text elements


mikejsavage:
edit: looking at the macro API, calling ConfigureOpenElement on a text element is just not valid code so there are No Bugs Here, but it would be nice to have a replacement for `CLAY_EXTEND_CONFIG_TEXT`.

---

Hi, it seems like clay doesn't forward `Clay_ElementDeclaration::userData` through to TEXT render commands:

```cpp
#include <stdio.h>
#define CLAY_IMPLEMENTATION
#include "clay.h"

int main( int argc, char ** argv ) {
    static char mem[ 10000000 ];
    Clay_Arena arena = Clay_CreateArenaWithCapacityAndMemory( sizeof( mem ), mem );
    Clay_Initialize( arena, { }, { } );

    auto measure_text = []( Clay_StringSlice text, Clay_TextElementConfig * config, void * user_data ) -> Clay_Dimensions {
        return { 5, 5 };
    };
    Clay_SetMeasureTextFunction( measure_text, NULL );

    Clay_SetLayoutDimensions( Clay_Dimensions { 100, 100 } );
    Clay_BeginLayout();

    Clay__OpenElement();
    Clay__ConfigureOpenElement( Clay_ElementDeclaration {
        .id = 1,
        .userData = ( void * ) 1,
    } );
    Clay__OpenTextElement( CLAY_STRING( "Hello" ), Clay__StoreTextElementConfig( Clay_TextElementConfig { } ) );
    Clay__ConfigureOpenElement( Clay_ElementDeclaration {
        .id = 2,
        .userData = ( void * ) 2,
    } );
    Clay__CloseElement();

    Clay_RenderCommandArray layout = Clay_EndLayout();
    for( int32_t i = 0; i < layout.length; i++ ) {
        const Clay_RenderCommand & command = layout.internalArray[ i ];
        const Clay_BoundingBox & bounds = command.boundingBox;
        printf( "%d %d,%d %dx%d %p\n", command.commandType, int( bounds.x ), int( bounds.y ), int( bounds.width ), int( bounds.height ), command.userData );
    }

    return 0;
}
```

I would expect this to produce a TEXT render command with the userData pointer set to 1 or 2, but instead clay emits a NONE render command (which I suppose may also be a bug, although it goes away if you don't configure the text element) with userData = 2 and a TEXT command with userData = NULL.

nicbarker:
Thanks for this one, looking into it now!

[...]

PR #274, opened 2025-02-17
[Core] Add a userData pointer to Clay_TextElementConfig

mikejsavage:
I guess this is the least effort fix for #271, feel free to close if you would prefer a different approach

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -357,6 +357,8 @@ typedef CLAY_PACKED_ENUM {
 
 // Controls various functionality related to text elements.
 typedef struct {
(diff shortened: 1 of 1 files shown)


nicbarker:
Thanks for this one, I'll just double check the struct alignment / padding and we can get this in 🙂

mikejsavage:
The struct is 4 bytes smaller on 32bit with `userData` at the top, makes no difference on 64bit

nicbarker:
Ah sorry, I thought I had merged this and it slipped my mind.

PR #275, opened 2025-02-18
Fix inverted condition for setting booleanWarnings.maxTextMeasureCacheExceeded

mizmar:


nicbarker:
Great catch, always easy to accidentally forget `!`. Thanks! 😁

PR #277, opened 2025-02-19
Adds Clay_GetPointerState

Orcolom:
Having pointer state access allows you to build flexible interactions and provide parity with what you can do with OnHover without needing OnHover. You can't use onHover if you want to return the result of your calculation. 

Currently you can get this through Clay_Context but many languages don't have access or need access to that internal data and use Clay_Context purely as a pointer/empty object.


OetkenPurveyorOfCode:
i tried the sokol renderer and it has no sclay_is_mouse_down() functionality (apart from using internal _sclay.mouse_down). I know for raylib and other backends it is probably possible to use IsMouseButtonDown(MOUSE_BUTTON_LEFT) or something, but at least  the sokol renderer did not expose this.

PR #278, opened 2025-02-19
[Compilers] Added DLL macro to support .dll building 

Orcolom:
To support DLL's c methods need to be explicitly exported. I have followed the WASM implementation as close as I could (see notes). These changes were needed to make the C# implementation work.

Some implementation notes:
- C isn't my strongest suit. So, its best this gets double checked this works correctly for non-dll exporting.
- I'm not entirely sure if the fallback `CLAY_DLL_EXPORT(null)` is correct in this case. 
- From my (imperfect) testing these had to be added to the declaration of the methods instead of the implementations like WASM can do, might again be me not fully understanding c macro syntax and that it was conflicting with the WASM macro somehow.
- The internal methods that exist for macros also have these so the implementor can recreate the macros.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -42,6 +42,12 @@
 #define CLAY_WASM_EXPORT(null)
 #endif
 
+#ifdef CLAY_DLL
+#define CLAY_DLL_EXPORT() __declspec(dllexport) __stdcall
+#else
+#define CLAY_DLL_EXPORT(null)
+#endif
+
 // Public Macro API ------------------------
 
 #define CLAY__MAX(x, y) (((x) > (y)) ? (x) : (y))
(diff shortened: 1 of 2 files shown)


clay.h
```diff
@@ -42,6 +42,12 @@
 #define CLAY_WASM_EXPORT(null)
 #endif
 
+#ifdef CLAY_DLL
+#define CLAY_DLL_EXPORT() __declspec(dllexport) __stdcall
```
eigenlenk, line 46: No need to call this as a function, You can drop the parenthesis and null argument.

nicbarker:
This looks good to me!

nicbarker:
Last thing we'll need is just a new section in the README to document the flag and what it does 🙂

nicbarker:
This looks good to me! Thanks 😁

Issue #276, opened 2025-02-19
CLAY_ID uniqueness


Roucou:
I'm struggling with the CLAY_ID concept, and watching the great sets of videos by [David Styrbjörn](https://www.youtube.com/c/DavidStyrbj%C3%B6rn), I'm not the only one.

Most examples you find on defining your own "components" do something like this:

```c
void button_element(const char*clay_id, Texture2D *texture) {
    CLAY({
            .id = CLAY_ID(clay_id),
            .image = {
                    .imageData = texture,
                    .sourceDimensions = { texture->width, texture->height }
            },
            .layout = { .sizing = {
                    .width = CLAY_SIZING_FIXED(texture->width),
                    .height = CLAY_SIZING_FIXED(texture->height),
            }}
    }) {}
}
```

But using clang, that doesn't even compile, it will complain about the CLAY_ID(clay_id) bit. A naive way around this would be:

```c
void button_element(Clay_ElementId *clay_id, Texture2D *texture) {
    CLAY({
            .id = *clay_id,
            .image = {
                    .imageData = texture,
                    .sourceDimensions = { texture->width, texture->height }
            },
            .layout = { .sizing = {
                    .width = CLAY_SIZING_FIXED(texture->width),
                    .height = CLAY_SIZING_FIXED(texture->height),
            }}
    }) {}
}
```

And then call this function as such:

```c
Clay_ElementId element_id = CLAY_ID("button1");
Texture_ID texture_id = TEX_BUTTON1;
Texture2D texture = textures[texture_id];
button_element(&element_id, &texture);
element_id = CLAY_ID("button2");
texture_id = TEX_BUTTON2;
texture = textures[texture_id];
button_element(&element_id, &texture);
```

But that somehow creates non-unique CLAY_ID's. A quick-fix would be to give each call its own local scope:

```c
{
  Clay_ElementId element_id = CLAY_ID("button1");
  Texture_ID texture_id = TEX_BUTTON1;
  Texture2D texture = textures[texture_id];
  button_element(&element_id, &texture);
}
{
  Clay_ElementId element_id = CLAY_ID("button2");
  Texture_ID texture_id = TEX_BUTTON2;
  Texture2D texture = textures[texture_id];
  button_element(&element_id, &texture);
}
```

Or, as David is doing, create static const char* id's in the global scope.

[...]

PR #285, opened 2025-02-25
[Macros] Add versions of the CLAY_ID macros that take Clay_String

FintasticMan:
These are named CLAY_SID*.

This is based on @Roucou's suggestion in https://github.com/nicbarker/clay/issues/276#issuecomment-2673611334, without changing the public API.

This allows you to re-use and compartmentalise IDs more easily.


nicbarker:
Great idea! I'll just get some small changes to the README in then we can merge it 👍

PR #294, opened 2025-03-03
[Core] Make `CLAY_MIN` and `CLAY_MAX` aware of type and double eval

rhodes-b:
I tried to run the docker tests it failed for me before this change so im not sure how to test it on my system. 
Normal build with examples also fails because I'm on a wayland only system so I cant build any of the demos due to a X11 dependency in raylib I believe

I was going to make an issue but this change is very easy so I decided to open a PR being clear you are absolutely free to close this for any reason.

refer to 
https://stackoverflow.com/questions/3437404/min-and-max-in-c
https://gcc.gnu.org/onlinedocs/gcc-4.9.2/gcc/Typeof.html#Typeof

It has a wrapper `{ }` which seems important (below is linked in the stack overflow)
https://gcc.gnu.org/onlinedocs/gcc/Statement-Exprs.html#Statement-Exprs
This fails to compile if you have `-Wpedantic` though even though it says you must use `__typeof__` in ISO C programs

Why is this useful?
The comment in the stack overflow, other gcc docs I linked can explain it better than I could. I just know this is typically the preferred method of defining a macro version of `min` and `max` in c

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -50,8 +50,8 @@
 
 // Public Macro API ------------------------
 
-#define CLAY__MAX(x, y) (((x) > (y)) ? (x) : (y))
-#define CLAY__MIN(x, y) (((x) < (y)) ? (x) : (y))
+#define CLAY__MAX(x, y) (__typeof__ (x) _x = (x); __typeof__ (y) _y = (y); ((_x) > (_y)) ? (_x) : (_y))
+#define CLAY__MIN(x, y) (__typeof__ (x) _x = (x); __typeof__ (y) _y = (y); ((_x) < (_y)) ? (_x) : (_y))
 
 #define CLAY_TEXT_CONFIG(...) Clay__StoreTextElementConfig(CLAY__CONFIG_WRAPPER(Clay_TextElementConfig, __VA_ARGS__))
 
--- a/clay.h
+++ b/clay.h
@@ -50,8 +50,8 @@
 
 // Public Macro API ------------------------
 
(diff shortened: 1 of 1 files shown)


rhodes-b:
looks like I may not have been using the proper commit naming convention I can redo this if I am not following your standards

FintasticMan:
So this is as a fix for the arguments being evaluated twice in the original macro? (Meaning that any function calls will be executed twice). Unfortunately both `__typeof__` and and statement expressions are GNU extensions and so don't work in MSVC. Given that these macros are internal macros (the double underscore means that it isn't part of the public API), I don't think that this is really necessary. Maybe it would be good to just add a comment above the definitions warning against using arguments with side-effects (or slow functions).

rhodes-b:
> So this is as a fix for the arguments being evaluated twice in the original macro? (Meaning that any function calls will be executed twice). Unfortunately both `__typeof__` and and statement expressions are GNU extensions and so don't work in MSVC. Given that these macros are internal macros (the double underscore means that it isn't part of the public API), I don't think that this is really necessary. Maybe it would be good to just add a comment above the definitions warning against using arguments with side-effects (or slow functions).

Completely slipped my mind I should have checked if Clay supported windows. 

I had also thought this can help the compiler evaluate invalid type comparisons but it seems neither `-Wall` or `-Werror` actually warn you about mixing types even with this at least with basic float and int calls. Like I said perfectly fine closing this. Double evaluation is typically only bad if you are passing expressions with side effects like a function call which I would assume most people arent doing anyways

PR #295, opened 2025-03-03
[Renderers/SDL2] Fix rounded corner border index

mizmar:


TimothyHoytBSME:
Oof. Good catch, I must've forgot to change that one when I copied the line. And it looks like there is something wrong with the PR checks, so @nicbarker may need to address that.

nicbarker:
Ah yes, unfortunately github deprecated some action runner things - will get them fixed up so we can merge this. Thanks for the good work! 👍

PR #298, opened 2025-03-03
[Renderers/Raylib] PERF: Fixed many allocations per frame for each string to render.

hexmaster111:
Updated `renderers/raylib/clay_renderer_raylib.c` to store the malloc'd buffer, so we dont need to malloc and free every string part.

Also added `Clay_Raylib_Close()` to free the buffer.


hexmaster111:
I just looked through the examples, and it looks like we are missing a few calls to `CloseWindow()` in raylib, If this is merged, ill update the raylib examples to call `Clay_Raylib_Close()` to de-init raylib, as well as free the stored buffer.

hexmaster111:
I updated the Raylib examples to all call the clay_raylib_close method.

nicbarker:
Thanks for the work here! Taking a look now 🙂

PR #300, opened 2025-03-04
[Documentation] Update README.md: it's gotten bigger

joshuahhh:


nicbarker:
Haha yes, good catch. Very tempting to do the classic programmer "spend 1 hour automating something that takes 5 seconds to update" and update that value dynamically...

PR #301, opened 2025-03-04
[Bindings/Odin] expose _OpenElement and _CloseElement

lzurbriggen:
I guess `_OpenElement` and `_CloseElement` are considered private because in C you will likely always use the provided macro. In Odin however there are no macros and exposing the functions allows creating nice component procedures, i.e.:

```odin
@(deferred_none = clay._CloseElement)
my_button :: proc(id: clay.ElementId) -> bool {
	clay._OpenElement()
	return clay.ConfigureOpenElement({layout = {padding = clay.PaddingAll(8)}})
}
```

Which can be used like this (without calling the higher-order `clay.UI` function):

```odin
if my_button(clay.ID("Btn")) {
	clay.Text("Hellope", clay.TextConfig({textColor = WHITE}))
}

/// instead of:

// my_button would return just a clay.ElementDeclaration here
if clay.UI()(my_button(clay.ID("Btn"))) {
	clay.Text("Hellope", clay.TextConfig({textColor = WHITE}))
}
```

### DIFF
--- a/bindings/odin/clay-odin/clay.odin
+++ b/bindings/odin/clay-odin/clay.odin
@@ -367,7 +367,7 @@ Context :: struct {} // opaque structure, only use as a pointer
 @(link_prefix = "Clay_", default_calling_convention = "c")
 foreign Clay {
 	_OpenElement :: proc() ---
-    _CloseElement :: proc() ---
+	_CloseElement :: proc() ---
 	MinMemorySize :: proc() -> u32 ---
 	CreateArenaWithCapacityAndMemory :: proc(capacity: u32, offset: [^]u8) -> Arena ---
 	SetPointerState :: proc(position: Vector2, pointerDown: bool) ---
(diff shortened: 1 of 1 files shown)


nicbarker:
Ah, this is awesome - this type of "custom component with user provided children" design isn't really possible in C without the user creating their own macros. From your example, it looks like we'd also need to make `ConfigureOpenElement` public too, right?

lzurbriggen:
@nicbarker 
Yeah, it's very nice to work with! I'm doing game UI stuff with Clay, super happy with this approach.

> it looks like we'd also need to make `ConfigureOpenElement` public too, right?

There is already a wrapper function for that one, so we don't have to touch it:

```odin
ConfigureOpenElement :: proc(config: ElementDeclaration) -> bool {
	_ConfigureOpenElement(config)
	return true
}
```

nicbarker:
Apologies for the delay, I thought I had merged this one. Thanks for the work!

Issue #283, opened 2025-02-23
Raylib example broke


EmmanuelMess:
The raylib example is broken and doesn't scroll

![Image](https://github.com/user-attachments/assets/83bd8f8d-90f5-487f-aab3-cc83b509e71b)

EmmanuelMess:
Okay, I was missing the font file, but I think it broke more than it should have (see the overwriting text at the bottom.)

hexmaster111:
I just had this issue too, I copied the resources folder to be next to the program file, and it looks like it should. Perhaps the build system should copy those files to the output folder for us?

EmmanuelMess:
> I just had this issue too, I copied the resources folder to be next to the program file, and it looks like it should. Perhaps the build system should copy those files to the output folder for us?

Even without the resources it shouldn't look like this. It should copy the resources when building.

hexmaster111:
If you take a look at the Raylib_MeasureText Method [here](https://github.com/nicbarker/clay/blob/02bce89d17e6afb020a34f500a2d78fb8dd3150a/renderers/raylib/clay_renderer_raylib.c#L84), it looks like when the font had failed to load, it returns ` Clay_Dimensions textSize = { 0 };` as the font size, so the text stacking up like that makes sense, as clay would have gotten a measurement of zero height. 

Perhaps adding a `IsFontValid()` call to where the font is loaded to check and then the application exits with a missing font message would help with user confusion?

EmmanuelMess:
> Perhaps adding a IsFontValid() call to where the font is loaded to check and then the application exits with a missing font message would help with user confusion?

yes, and the text is visible, so the return value of 0 is wrong.

hexmaster111:
> so the return value of 0 is wrong

True, I think i'll work to update the render to measure the default RayLib font then if a loaded font is invalid. Thanks for pointing that out to me!

hexmaster111:
If you check out #305

Now we can use the default raylib font and it looks kinda alright. 

![Image](https://github.com/user-attachments/assets/506b6ec5-22af-459a-b6bf-b524554fde64)

turns into 

![Image](https://github.com/user-attachments/assets/cdc59201-b7fe-4f10-9f3e-a6710633146f)

PR #305, opened 2025-03-04
[Renderers/Raylib] Use RayLib Default font if user font failed to load or was not specified.

hexmaster111:


hexmaster111:
Fixes issue #283

nicbarker:
Wow this is awesome, I have wanted to do something like this for ages, great job!

PR #306, opened 2025-03-04
[Bindings/Odin] fix CreateArenaWithCapacityAndMemory capacity type

laytan:
Wrong since 5571c00


nicbarker:
Thanks for this pickup!

PR #307, opened 2025-03-04
[Documentation] fix example in README

RicoP:
example has one too many `}`. Took me a while to find this.


nicbarker:
Ah, one snuck through 😅 Thanks for catching this!

PR #313, opened 2025-03-06
[Renderers/Raylib] Support unicode in Raylib_MeasureText

ysoftware:
When running my code in Xcode with all sorts of sanitizers, I got the "Heap buffer overflow" error in a string containing unicode characters.

I noticed that the code did not support unicode correctly, so I copied an implementation of this loop from my own text renderer. It uses Raylib's `GetCodepoint` and `GetGlyphIndex` functions and makes no assumptions about the string structure.

My application uses a monospaced font, so I did not have a chance to test this other than running the 2 example raylib projects, which looked correct.

### DIFF
--- a/renderers/raylib/clay_renderer_raylib.c
+++ b/renderers/raylib/clay_renderer_raylib.c
@@ -99,16 +99,22 @@ static inline Clay_Dimensions Raylib_MeasureText(Clay_StringSlice text, Clay_Tex
 
     float scaleFactor = config->fontSize/(float)fontToUse.baseSize;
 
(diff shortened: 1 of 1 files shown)


igadmg:
Why that is even done by hand? Why not use raylibs MeasureTextEx function?

ysoftware:
Good point. I didn't know that existed :D
I can apply that function instead in scope of this PR a bit later.

ysoftware:
I have tried to use MeasureTextEx and it requires a zero-terminated string, which is not what we're getting with `Clay_StringSlice`. I think I will rather stand by the current implementation.

PR #316, opened 2025-03-07
[Compilers] Fixed SIMD related compile error on some ARM compilers

emoon:
When compiling for arm using GCC it would complain about the following code because the argument wasn't correct for `vminq_u32` I have now added a cast (that just reinterprets the type so no overhead) to fix the compile error.


codewithpapakwame:
cool huhh

nicbarker:
Nice, tested and it looks good to me!

PR #317, opened 2025-03-08
[Compilers] Fix integer truncation warnings with explicit casts

Nelarius:
Hey Nic,

Very cool project, having fun experimenting with Clay in a UI project 🙂 

Fixes some integer truncation warnings, which prevent using clay with a bunch of compiler warnings enabled, such as `-Wall -Wextra -Wpedantic -Werror`:
1.  `Clay__int32_tArray_GetValue` expects an index of type `int32_t`.
2. `Clay_Arena::nextAllocation` is a uintptr_t, explicitly truncate the return value.


nicbarker:
Thank you! Will test these out locally then get them merged 👍

nicbarker:
Looks good to me!

PR #319, opened 2025-03-08
Support passing declaration by pointer as well

emoon:
This change adds `Clay__ConfigureOpenElementPtr` which allows passing in `Clay_ElementDeclaration` by pointer instead of by value. This allows the user to have pre-built `ElementDeclaration`


nicbarker:
Great, thanks for this change! Have been meaning to get it in for a while. I'll have a think about how best to expose this via a new macro in C.

PR #320, opened 2025-03-09
[Renderers/SDL2] Enable live resizing of layout during window resize in SDL2

shakkar23:
Added calling the rendering functions while the window is being resized!

A feature that you had said was an issue in your video!

The SDL_AddEventWatch function is for some reason hard to find and has very little examples on how to use online.

Hope this is helpful!


nicbarker:
Wow, this works perfectly! Thank you and great job 😁

PR #327, opened 2025-03-15
[Renderers/SDL2] Added explicit include of math.h in SDL2 renderer

Emerald-Ruby:
GCC Complaining about not including math.h still runs without the include, but a wall of "use math.h" is printed

### DIFF
--- a/renderers/SDL2/clay_renderer_SDL2.c
+++ b/renderers/SDL2/clay_renderer_SDL2.c
@@ -3,6 +3,7 @@
 #include <SDL_ttf.h>
 #include <SDL_image.h>
 #include <stdio.h>
+#include <math.h>
 
 #ifndef M_PI
     #define M_PI 3.14159


FintasticMan:
One of the things Clay aims for is to not depend on the C standard library, including math.h. Could you show me what the compiler warnings are? I haven't noticed them myself.

Emerald-Ruby:
Nothing critical, honestly you can reject it, it's just a tiny thing that doesn't effect it, but my mingw64 GCC compiles the SDL2 renderer with an implicit math.h, I am on windows too

image

nicbarker:
Hey @Emerald-Ruby - you're absolutely right, there is an implicit usage of `math.h` functions in the SDL2 renderer. We're totally fine with having standard lib usage in the renderers and examples outside of the core `clay.h`, thanks for catching this!

PR #336, opened 2025-03-21
Add struct names to public structs

### LINKED ISSUE #318: Declare public types with tags to allow forward declaration
**Problem**

In Clay, the public API structs are declared without tags:

```c
typedef struct {
    // fields...
} Clay_RenderCommandArray;
```

This means that, at least to my knowledge, the types can't be forward-declared when defining things in a header for e.g. a render backend:

```c
#pragma once

// NOTE: I must include clay.h for Clay_RenderCommandArray
#include <clay.h>

struct clay_gpu_context;
struct gpu_render_pass;

void clay_gpu_render_commands(
    struct clay_gpu_context*, struct gpu_render_pass*, Clay_RenderCommandArray* commands,
    vec2s frame_size);
```

To be clear, this is a somewhat pedantic request, but it's nice to be able to keep headers as free of includes as possible 🙂 

**Proposal**

Declare public types as

```c
typedef struct Clay_RenderCommandArray {
    // fields...
} Clay_RenderCommandArray;
```

to allow forward-declaring the tag names in our headers.

Would be happy to make a pr if this is something deemed useful.

Nelarius:
Closes #318 .
* Adds struct names to all public structs. This makes it possible to forward-declare the public types without including `clay.h`:
```c
struct Clay_RenderCommandArray;
struct my_clay_context;

void my_clay_renderer(struct my_clay_context*, struct Clay_RenderCommandArray const*); 
```
* Leaves enums untouched for simplicity. Only reason to add names to enums would be for consistency with the other type declarations.
* All examples and bindings should compile with these changes.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -184,7 +184,7 @@ extern "C" {
 
 // Note: Clay_String is not guaranteed to be null terminated. It may be if created from a literal C string,
 // but it is also used to represent slices.
-typedef struct {
+typedef struct Clay_String {
     // Set this boolean to true if the char* data underlying this string will live for the entire lifetime of the program.
     // This will automatically be set for strings created with CLAY_STRING, as the macro requires a string literal.
     bool isStaticallyAllocated;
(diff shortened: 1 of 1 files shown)


nicbarker:
Sorry for the delay in getting to this one, thank you for the contribution! 🙂

PR #338, opened 2025-03-22
[Core] Guard against hashmap item null dereference

igadmg:
mapItem->boundingBox is accessed before mapItem null check.


nicbarker:
Nice catch, thanks for this one!

PR #340, opened 2025-03-22
Add interface target to be used with Meson's CMake module

bonk256:
Add an INTERFACE library target so that Clay can be used as a subproject within Meson, instead of having to be hard linked within the sources / CFLAGS.

The target won't produce any file artifacts while allowing you to register of a dependency with `<clay_cmake_subproject>.dependency('headers')`

This helps modularize Meson projects which depend on Clay and minimizes bloat by not requiring users to create an additional directory for dependencies.


ketexon:
I would like this also so that clay could be used as a CMake subproject, so I think the name of the target should be something more specific like `clay`.

PR #344, opened 2025-03-23
[Renderers/WinGDI] Working on Win32 GDI renderer and example

Philosoph228:


Philosoph228:
I assume there's no tests for win32 for now. Idk should them be? Like setting up Wine and cross-compiling binaries with MinGW toolchain.

nicbarker:
@Philosoph228 I don't think we need tests just for the moment, I'm unlikely to be modifying the gdi renderer any time soon 🙂

nicbarker:
This is looking awesome! Last thing left to tidy up would be just switching the `USE_INTRINSICS` define over to use the `CLAY_DISABLE_SIMD` define instead, as that's currently the flag that enables / disables intrinsics in clay.h, and it would be good to have them both use the same define 🙂

Philosoph228:
Good! I'll make that work with `CLAY_DISABLE_SIMD` definition. I'm also going to double-check the resource management and polish the code a bit - I don't really like that multi-line CreateFont() call.

PR #345, opened 2025-03-24
[Cmake] Basic CMake support for easier import into CMake projects

Qualadia:
This PR will allow to use library in a manner provided below:
```
set(CLAY_INCLUDE_ALL_EXAMPLES OFF)
FetchContent_Declare(
	CLAY
	GIT_REPOSITORY "https://github.com/Qualadia/clay.git"
	GIT_TAG main
	GIT_SHALLOW TRUE
	GIT_PROGRESS TRUE
)
FetchContent_MakeAvailable(CLAY)

target_link_libraries(${PROJECT_NAME} PRIVATE
	clay
)
```


nicbarker:
Thank you for the help with this! I'm not very familiar with CMake myself so appreciate it when other people know the magic incantations 🙂

nicbarker:
Unfortunately this broke my local dev environment on OSX / Clang / CLion, I will need to take a better look at why, but for now I've reverted the change 🥲

PR #348, opened 2025-03-25
Fix CLAY__ELEMENT_DEFINITION_LATCH overflow in CLAY macro if 256 loops end at the same time

PiggybankStudios:
Incrementing the `CLAY__ELEMENT_DEFINITION_LATCH` in the post-for-loop logic can cause undesired effects if it gets incremented 256 times in a row. This only happens if 256 `CLAY` scopes end in a row without another CLAY scope starting.

To fix this, we can simply set `CLAY__ELEMENT_DEFINITION_LATCH=1` rather than incrementing it.


PiggybankStudios:
Woops this pull request contains 2 commits when it should only contain one. I will close this one and create another one with only the one commit

PR #349, opened 2025-03-25
[Core] Fix CLAY__ELEMENT_DEFINITION_LATCH overflow in CLAY macro if 256 loops end at the same time

PiggybankStudios:
Incrementing the `CLAY__ELEMENT_DEFINITION_LATCH` in the post-for-loop logic can cause undesired effects if it gets incremented 256 times in a row. This only happens if 256 `CLAY` scopes end in a row without another `CLAY` scope starting.

To fix this, we can simply set `CLAY__ELEMENT_DEFINITION_LATCH=1` rather than incrementing it.


nicbarker:
Great catch! You must be building something pretty big if you got to 256 nested scopes 🙂

PiggybankStudios:
Full disclaimer, I did not actually run into a use case here. I was discussing for loops being used as a sort of "defer" like behavior with some coworkers and mentioned how Clay does this in its CLAY macro. But when I looked at the code I was a bit confused by the single static uint8_t being used for all loops and being incremented. Thought it might be that something was using that to track depth in the hierarchy but didn't find any other references. After contemplating if this approach really works I realized there was a tiny edge case (as described), not a common one by any means, but also not hard to fix either so I figured it was worth making the pull request

nicbarker:
Ah I see, well thanks for the keen eyes 🙂

PR #350, opened 2025-03-25
[Bindings/Odin] Remove field hashStringContents in odin bindings

ellie-but-backwards:
PR  #335 removed this field in clay.h but the bindings didn't reflect that change


nicbarker:
Ah thank you, oversight on my part 🙂

PR #357, opened 2025-03-28
[Renderers/SFML2] Added initial SFML2 support

matthew-nagy:
Add support for a renderer for SFML 2. SFML 2 is a cpp library, so I'm using cpp for this, rather than pure c.


renderers/SFML2/clay_renderer_SFML2.hpp
eXpl0it3r, line 1: Should this header have an include guard?

PR #362, opened 2025-04-01
Resizeable Images + Example Image Loader for Win32 GDI

kylebueche:
I added some support for windows bitmap rendering that uses .image attributes as pointers to HBITMAP. I also included some example files which are just copies of the existing Win GDI example, modified to include images. Since the examples are just a direct copy of the video example, but with images added, they don't have to make the final cut. It's just to show it in action, and the main change I want to contribute is that image rendering is now supported via bitmaps. The user can load or generate bitmaps however they like.

- Currently supports expanding to fill the bounding box maximally while preserving aspect ratio.
- Does not yet support corner rounding, tinting, or alpha channel transparency.

Open to any revisions!


nicbarker:
Wow, thank you for all the work here! I will take a look when I'm back at my PC 🙂

Issue #337, opened 2025-03-21
Corner Radius behaviour quirks


Emerald-Ruby:
### The problem : 
  I am trying to make a container with a rounded top right corner, but no other. Whenever I set the top left, it is applied to the whole object, and if everything is set to 15 and only the top left is set to 0, no radius is applied.

### Possible points of error :
  I have ran surface level debugging, the renderer correctly see the corner radius as it's intended values (image 1), and when printing the types the SDL2 receives, all containers are rectangles (image 2), and the code to deal with that renderer command only checks the top left corner, with my limited knowledge of how the renderer commands work and how they are meant to be interpreted, I believe the error is in either the clay.h when making the command it incorrectly assigns the container a rectangle shape OR the SDL2 renderer incorrectly deals with the container

  from my peaks at the SDL2 renderer, it does seem like clay.h is producing the wrong command, since the final case in the SDL2 renderer looks like it would handle the corners correctly (image 3)

### Conclusion :
Of course this could be an intentional thing left to be dealt with later, or could bug, maybe it is a concern or maybe it is not, maybe I have programmed something wrong, but I have gone onto the video example code, and used just {0, 8, 8, 8} and also what I have shown and still the same results (image 4 and 5)

**Image 1:**
![Image](https://github.com/user-attachments/assets/820836ea-3c87-4913-82c8-c6368027e658)

**Image 2:**
![Image](https://github.com/user-attachments/assets/981a4e5e-cbe2-4a88-b05a-d831eacc19de)

**Image 3:**
![Image](https://github.com/user-attachments/assets/5ef8db7d-47eb-4e9a-b52f-d576b6d58850)

**image 4:**
![Image](https://github.com/user-attachments/assets/a8be108f-abba-403b-932f-e760565cc341)

**image 5:**
![Image](https://github.com/user-attachments/assets/64558c0f-64c3-4c77-b55c-c99d7c97b5a0)

Emerald-Ruby:
Small update, the problem still persists in the skeleton example as above, however while adding UI elements with borders it fixed itself. It is a very odd behaviour, and I haven't investigated different configurations of the code, but it does lead me to believe if the border isn't set but the corner radius is, it is threated like a rectangle object, this is only my speculation though.

![Image](https://github.com/user-attachments/assets/43708f0b-5029-411a-b54e-c114be346264)
![Image](https://github.com/user-attachments/assets/7006202c-5f65-4ab6-b21f-0210694cc38b)

PR #373, opened 2025-04-08
[Renderers/Sokol] Sokol renderer & examples

nkorth:
Working sokol renderer using sokol_fontstash for text (no image support yet.) I tried to follow the sokol code style, including minimal dependencies.

This renderer properly handles unequal cornerRadius, as was brought up in issue #337 (for SDL2.) Not sure if it's really worth the increased complexity in every renderer, but I thought I'd try it for this one.

Screenshot 2025-04-08 at 3 34 14 PM


nicbarker:
Just tried this out and it looks incredible. Thanks so much for the hard work!

nkorth:
Not at my computer right now but I bet there's something in https://github.com/floooh/cimgui-sokol-starterkit/blob/main/CMakeLists.txt which I should have left in the CMakeLists. For example I'm pretty sure the windows main stub is supposed to come from sokol_app automatically

nicbarker:
@nkorth Ah yes, I see it. Thanks for the link, will try it out now 🙂

Issue #359, opened 2025-03-29
[Core] Seperate Clip config & scroll handling improvements


Muzz:
Right now vertical scroll is lacking some needed custom functionality. Atm, it has momentum, which i don't want in my scrolling, and there is no way to disable it.  

While it makes for an easy setup first time through, once you get deep into development and you want to extend off the scrolling it gets in the way, so I would suggest that we have an option to turn off the automatic handling, and give a function to get the current scroll, and set the scroll. All this would make it far easier to add proper UI to handle scrolling. 

Also, there doesn't seem to be a way to turn on scissor, without scrolling being on. 

Thanks!

nicbarker:
Definitely agree here, next up on the list is splitting `clip` into a seperate config item, it will be roughly this shape:
```C
struct ClipConfig {
  bool enabled;
  Vector2 contentOffset; // x,y offset for clip contents, used for handling scrolling
}
```

I'm thinking we'll split out scrolling into something seperate, with the view to eventually move Clay's scroll handling implementation into a different head along with the other interactions. In the short term it will look something like:

```C
CLAY({
  .clip = {
    .enabled = true,
    .offset = Clay_UpdateScrollContainer(/* params like mouse wheel state etc go here */)
  }
});
```

To me this feels much more in line with "immediate mode" style - if you handle scrolling yourself, you would just pass the current `.offset` to the clip config each frame.

Muzz:
Those solutions both sound ideal to me!

what number would that offset be? A percentage would be easiest, as then it's not reliant on the container size, but also means it's harder to calc the offset. 

Solid thinking.

nicbarker:
This has been completed by https://github.com/nicbarker/clay/pull/376 merging 🙂

PR #376, opened 2025-04-10
[Core] Replace .scroll config with .clip

nicbarker:
Addresses https://github.com/nicbarker/clay/issues/359

This PR replaces the `.scroll` element configuration field with a more generic `.clip` option.
Clip has the same `.horizontal` and `.vertical` booleans, but also provides a `Vector2 .childOffset` field, which is used to offset the positions of all child elements - essentially the building block for implementing scrolling.

Clay still handles scrolling for you in exactly the same way, with the only difference being that you now need to provide the scroll offset for each element:

```C
// Before
CLAY({
  .scroll = { .vertical = true }
});

// After
CLAY({
  .clip = { .vertical = true, .childOffset = GetScrollOffset() }
})
```

There are two main motivations behind this PR:

- Previously, it was impossible to enable clipping on an element _without_ enabling scrolling (i.e. overflow:hidden from CSS). This is now possible.
- This step takes us closer to our goal of splitting the code that handles interactions into a [seperate header](https://github.com/nicbarker/clay/issues/231).

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,18 +178,10 @@ For help starting out or to discuss clay, considering joining [the discord serve
     - [Clay_GetScrollContainerData](#clay_getscrollcontainerdata)
     - [Clay_GetElementId](#clay_getelementid)
   - [Element Macros](#element-macros)
-    - [CLAY](#clay-1)
+    - [CLAY](#clay)
     - [CLAY_ID](#clay_id)
     - [CLAY_IDI](#clay_idi)
-    - [CLAY_LAYOUT](#clay_layout)
-    - [CLAY_RECTANGLE](#clay_rectangle)
-    - [CLAY_TEXT](#clay_text)
-    - [CLAY_IMAGE](#clay_image)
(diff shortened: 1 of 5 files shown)


Riquelme02:
I'm a little confused as to whether I'm misunderstanding these changes
or if the code is not working correctly.

>Clay still handles scrolling for you in exactly the same way

I thought that by specifying `childOffset = { y = -50.0}', for example,
that would simply offset the children of the clipping parent by 50 pixels,
ie, scroll the element down by that amount, since that's really all you need
to implement scrolling assuming the user is already handling the scrolling
offset themselves.  But setting that value in the element declaration, without
modifying any state related to the scroll container data, doesn't have any 
effect: the clipping works as expected, but the offset doesn't.

nicbarker:
@Riquelme02 thank you so much for reporting this, you're totally right, I made a small mistake in the implementation, it's fixed here: https://github.com/nicbarker/clay/commit/90b45f059cb2dc2848045b787dbd6f1999ac1ab5

PR #379, opened 2025-04-13
[Renderers/Raylib] Add explicit type cast for malloc

boringlily:
Make the type cast explicit to avoid using "-fpermissive".


nicbarker:
Looks good to me, thanks!

PR #389, opened 2025-04-24
[Core] Add `Clay_GetPointerOverIds` function to the public API.

tritao:
This adds a new `Clay_GetPointerOverIds` function to the public API that returns the array of element IDs that the pointer is currently over.

This also required to add a new public `Clay_ElementIdArray`, and to modify the naming of some internal array definitions to match.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -238,6 +238,14 @@ typedef struct {
     Clay_String stringId; // The string id to hash.
 } Clay_ElementId;
 
+// A sized array of Clay_ElementId.
+typedef struct
+{
+    int32_t capacity;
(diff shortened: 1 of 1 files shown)


nicbarker:
Thank you for the work here! Have been meaning to get to this one for some time, so it's great to cross it off the list 🙂

nicbarker:
Just a note to myself for later, when we move out interaction handling ala https://github.com/nicbarker/clay/issues/231, this function will be likely renamed something like `Clay_HitTest`, which will take the coords of the screen point to test and return the list of overlapped element ids 🙂

PR #391, opened 2025-04-24
[Core] Update `Clay__OpenTextElement` to return element id of text element.

tritao:
This updates `Clay__OpenTextElement` function to return element id of text element.


nicbarker:
Thanks for opening this one, something like this is obviously needed if we want to make it less painful to build editable text fields. Just a high level Q - what do you think about instead changing the signature of `CLAY_TEXT` (i.e. `Clay__OpenTextElement` underneath) to optionally accept an ID?
The reason text elements don't take / expose an ID at the moment is because I didn't figure out the edge cases around floating elements - if you attach a floating element to a text element via `.attachTo = element id`, where does it attach to? The enclosing axis aligned bounding box I suppose? Will that create unexpected behaviour when interacting with text wrapping and alignment?
Regardless, the utility obviously outweighs the possible edge cases at the moment, and I'm personally leaning towards just changing `CLAY_TEXT` in some way to be able to take an ID like `CLAY` does, but I would like to hear your 2c as you're actively using this at the moment 🙂

nicbarker:
I also just remembered that text elements having IDs is problematic when interacting with deferred rendering systems - at the moment each of our wrapped lines of text is emitted as a seperate render command with a unique ID that is derived from the base text element ID, so that might also produce unexpected results.

PR #395, opened 2025-05-01
[Bindings/Odin] Odin Raylib renderer rewrite

rats159:
Handful of changes to make the example Raylib renderer in Odin feel more idiomatic, and deduplicate some code


nicbarker:
Thank you for this work! I'll take a look and then we can get it merged 🙂

nicbarker:
Just ran this locally and it works great, well done!

PR #397, opened 2025-05-02
[Bindings/Odin] Updated odin bindings with new clip config

A1029384756:
updates the odin bindings to use the new clipping config that separates clipping and scrolling. introduced this in my application first to test


nicbarker:
Great work, thanks for moving on this so quickly! 😁

PR #402, opened 2025-05-03
[Renderers/SDL3] Use `SDL_Texture` instead of `SDL_Surface` for images

linkdd:
# Context

At the moment, the SDL3 renderer image feature is very inefficient. It expects in `imageData` an `SDL_Surface*`, and then creates an `SDL_Texture*` that is immediately destroyed.

Creating an `SDL_Texture*` every single frame is very inefficient and can kill the framerate. This operation uploads the image data to the GPU, which is a rather slow operation. You usually want to do that only once.

# Changes

 - Modify the image rendering feature so that `imageData` expects an `SDL_Texture*` instead of a `SDL_Surface*`
 - Update the SDL3 example to load the sample image as a texture instead of a surface
 - Add an example of how to display an image using clay that is not present in the shared layout

### DIFF
--- a/examples/SDL3-simple-demo/main.c
+++ b/examples/SDL3-simple-demo/main.c
@@ -24,6 +24,7 @@ typedef struct app_state {
 } AppState;
 
 SDL_Texture *sample_image;
+bool show_demo = true;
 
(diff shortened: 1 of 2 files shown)


nicbarker:
Great work, thank you for this contribution!

PR #403, opened 2025-05-05
Fix README: use correct anchor to Clay_CustomElementConfig

dgellow:
Tiny fix to the README, it seems the anchor to the custom element config section was incorrect


nicbarker:
Thanks for this fix! 🙂

PR #404, opened 2025-05-05
[Renderers/Playdate] Playdate console example

mattahj:
This PR adds a renderer and an example for the (Playdate console)[https://play.date/]. This is not ready yet but I'm creating this as a draft in case I have done something incredibly wrong that someone might notice early on before I get too far with it 😅 

The CMakeLists and playdate boilerplate C code in the example is pretty much the same as the 'Hello World' example from the playdate SDK examples, which I think is a good thing as it would be familiar to playdate devs that might want to try out Clay.

The remaining work before I take this out of draft would be:

- [x] Test the image rendering properly
- [x] Test multiple fonts. Playdate wants font assets to be created at specific sizes, so I would need to create some font assets for that. At the moment the example only uses the system font which is bundled with the Playdate, so no assets needed!
- [x] Support different border radius and thickness for each side of a rectangle

Clay compiled just fine for the Playdate! Thanks for such an awesome library!

Demo running on the simulator:

simulator

Demo running on the hardware:

https://github.com/user-attachments/assets/eed9b088-42f4-4286-a4a5-c6bb7aba655b


nicbarker:
This is so cool! Amazing work!! 😁
Will be glad to merge this once you're happy with it.

mattahj:
I got some time today and implemented the corner radius rendering properly
image

Hopefully tomorrow I can get the other parts done!

mattahj:
Turns out that there is more than 1 system font available on the playdate, so I could add more than 1 font to the example without having to add any font assets to the project 👌 
image

mattahj:
I decided not to do the color to dither pattern thing. It is doable using `playdate->graphics->setColorToPattern`, but users might have different opinions about what kinds of dither patterns should be used, so I will leave it up to them to implement using custom element config data or something like that

nicbarker:
Sorry for the delay in getting to this, it's looking so good! Thank you for this awesome contribution 😁

nicbarker:
Tried it out locally and it worked perfectly the first attempt with CMake 🎉

Screenshot 2025-05-19 at 11 45 52 am

PR #407, opened 2025-05-08
[Bindings/Odin] Add missing border macros to Odin bindings

Kelimion:


nicbarker:
Thanks for this one! Wasn't sure if people were using them, but if you find them useful, I'm happy to have them in the bindings 🙂

PR #413, opened 2025-05-13
[Core] Add Clay_FloatingClipToElement

pdoane:
This enables using Clay_FloatingAttachToElement without inheriting the clip context of the attached parent

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -445,6 +445,14 @@ typedef CLAY_PACKED_ENUM {
     CLAY_ATTACH_TO_ROOT,
 } Clay_FloatingAttachToElement;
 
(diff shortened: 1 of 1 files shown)


pdoane:
I added all the missing data to the debug view

nicbarker:
Now that I'm thinking about it, I actually wonder if not inheriting clipping is actually the better default behaviour, I have a feeling more people expect floating to be "over the top of everything else" by default, with inheriting of the clip as the niche case 🤔

pdoane:
I agree - not clipping as default is great, but I didn't want to break anything here.  Thanks for changing it!

PR #416, opened 2025-05-18
[Renderers/Cairo] Fix cairo renderer and example

zordythezordan:
There was some changes in clay function calls, so I changed argument types in cairo renderer + had to add ".chars", ".length" and ".isStaticallyAllocated" to Clay_String toTerminate declarations


nicbarker:
Thank you for catching these, it's much appreciated! I need to get the CI fixed up so these will actually be caught...

PR #418, opened 2025-05-22
[Repo]: Add Nix Flakes support

h4rldev:
I've added a `flake.nix` to this repository, so that developers using the Nix ecosystem can quickly pull the header and include it globally by just adding the link to the repository. 

It exports a package, a development shell environment, and a format checker. (mostly for the flake itself, but it checks the C code within the `examples` and `renderers`)

It's been tested to be working, at least on NixOS, testing for other Nix ecosystems like nix-darwin, may be needed, and testing on other platforms, such as ARM


h4rldev:
Oh, I forgot to mention, there may be dependencies needed for each example and build, but during my test builds, I haven't spotted anything wrong!

h4rldev:
It seems I forgot to include the .clang-format file the flake depends on, though I think you should review it's config regarding that so that macros aren't mangled and properly formatted, as I haven't really looked into it, I'll just remove the c formatting from the formatting command entirely, but that's your decision

h4rldev:
I'm pretty sure you can seper the flake packages to be both releases and branch releases :P

h4rldev:
Okay there we go, it's a bit ugly with the force pushes, but I have squashed the commits

SyntaxError2505:
Hello! I considered using clay for a project on nix and looked up if there was a flake. Found this pr. Sounds like it isn't really being worked on but I tried it out and it worked on my system. So is this getting merged sometime and would you consider publishing to nixpkgs?

h4rldev:
> Hello! I considered using clay for a project on nix and looked up if there was a flake. Found this pr. Sounds like it isn't really being worked on but I tried it out and it worked on my system. So is this getting merged sometime and would you consider publishing to nixpkgs?

I could perhaps enhance it further, this pr has been dangling for a while, with mostly no feedback, that I forgot about it.

I'll see if I can submit it as a package on the nix package repo sometime in the near future if I have time

SyntaxError2505:
Before nixpkgs, I'd prefer the PR to be merged so it doesn't become unmaintained and orphaned

Issue #409, opened 2025-05-10
[QUESTION] .backgroundColor to TEXT and BORDER


AlexB52:
Hi, 

I'm currently working on a TUI renderer for Clay via termbox2. Rendering any characters (borders included) often requires providing a background color.

Questions:

1. It doesn't seem intuitive for texts and borders to have a background color, but is there anything that really prevents these elements from having one? I have tinkered with something like the code below to get by. Is that something that could be considered in Clay?
2. As an alternative, is it reasonable to rely on the order of the Clay_RenderCommandArray to set a background color variable from a parent so that its children nodes can access it?
3. Is there another way for a child node to access its parent background color with the current Clay implementation?

Thank you

```diff
index 27de124..e068cbe 100644
--- a/clay.h
+++ b/clay.h
@@ -368,6 +368,7 @@ typedef struct {
     void *userData;
     // The RGBA color of the font to render, conventionally specified as 0-255.
     Clay_Color textColor;
+    Clay_Color backgroundColor;
     // An integer transparently passed to Clay_MeasureText to identify the font to use.
     // The debug view will pass fontId = 0 for its internal text.
     uint16_t fontId;
@@ -516,6 +517,7 @@ typedef struct {
 // Controls settings related to element borders.
 typedef struct {
     Clay_Color color; // Controls the color of all borders with width > 0. Conventionally represented as 0-255, but interpretation is up to the renderer.
+    Clay_Color backgroundColor;
     Clay_BorderWidth width; // Controls the widths of individual borders. At least one of these should be > 0 for a BORDER render command to be generated.
 } Clay_BorderElementConfig;
 
@@ -530,6 +532,7 @@ typedef struct {
     Clay_StringSlice stringContents;
     // Conventionally represented as 0-255 for each channel, but interpretation is up to the renderer.
     Clay_Color textColor;
+    Clay_Color backgroundColor;
     // An integer representing the font to use to render this text, transparently passed through from the text declaration.
     uint16_t fontId;
     uint16_t fontSize;
@@ -586,6 +589,7 @@ typedef struct {
     // Controls a shared color for all this element's borders.
     // Conventionally represented as 0-255 for each channel, but interpretation is up to the renderer.
     Clay_Color color;
+    Clay_Color backgroundColor;
     // Specifies the "radius", or corner rounding of this border element.
     // The rounding is determined by drawing a circle inset into the element corner by (radius, radius) pixels.
     Clay_CornerRadius cornerRadius;
[...]

PR #419, opened 2025-05-23
[Renderers/termbox2] Termbox2 renderer & examples

Mivirl:
Implemented a terminal renderer using [termbox2](https://github.com/termbox/termbox2)

What works:
- Multiple color modes:
  - Normal
  - 256 color
  - Grayscale
  - Truecolor
  - No_color
- Emulated transparency (for background colors only, not text)
- Mouse clicks, scrolling, and hover
- Supports the [NO_COLOR environment variable](https://no-color.org/)
- Some drawing settings can be changed by setting environment variables:
  - CLAY_TB_COLOR_MODE
    - NORMAL
    - 256
    - 216
    - GRAYSCALE
    - TRUECOLOR
    - NOCOLOR
  - CLAY_TB_BORDER_CHARS
    - DEFAULT
    - ASCII
    - UNICODE
    - BLANK
    - NONE
  - CLAY_TB_TRANSPARENCY
    - 1
    - 0
  - CLAY_TB_CELL_PIXELS
    - widthxheight

Limitations:
- Rounded corners are not supported
- Images are replaced with a placeholder pattern
- Borders are drawn on top of text
- Elements sometimes have different sizes based on their x,y position
  - Because Clay uses smaller pixels instead of large cells, some elements will be snapped to the nearest cell position, resulting in the same element being rendered across different numbers of cells depending on it's position

Possible future improvements:
- Add rounded rectangle support (maybe use a flag for this, since it may not make sense on a terminal)
- Add image support by adapting the logic from [derasterize](https://github.com/csdvrx/derasterize) or a similar project
- Have clay send border drawing commands before text drawing commands so text is drawn on top.
- Add a grid-snapping feature to clay to snap all measurements to the nearest cell size so dimensions are more consistent.

Screenshots:

Truecolor mode:

clay_tb_truecolor_border_round

Truecolor mode with transparency:

clay_tb_truecolor_border_round_transparent

Truecolor mode with transparency (debug window open):

clay_tb_truecolor_border_round_transparent_debug

256 color mode:

clay_tb_256_border_round

256 color mode with transparency:

clay_tb_256_border_round_transparent

Grayscale mode:

clay_tb_grayscale_border_round

Normal color mode:

clay_tb_normal_border_round

No color mode:

clay_tb_nocolor_border_round

Truecolor mode with minimum border width (borders are drawn over text):

clay_tb_truecolor_border_minimum


nicbarker:
Wow this looks so cool! Thank you for this contribution, I'll try to get it running on my machine and let you know how it goes 😁

renderers/termbox2/clay_renderer_termbox2.c
```diff
@@ -0,0 +1,1190 @@
...
+                    for (int x = box_begin_x; x < box_end_x;) {
+                        uint32_t ch = ' ';
+                        if (i < text->length) {
+                            int codepoint_length = tb_utf8_char_to_unicode(&ch, text->chars + i);
+                            if (0 > codepoint_length) {
+                                clay_tb_assert(false, "Invalid utf8");
+                            }
+                            i += codepoint_length;
+
+                            uintattr_t color_tb_bg = clay_tb_color_buffer_termbox_get(x, y);
```
AlexB52, line 1107: That's awesome! I've been working on a renderer too for tembox2 but at this point yours looks more complete. 

I know that termbox2 buffer is a private API and you implemented a bespoke color buffer to access background colors for text. I took the approach of adding background color to borders and text instead with the idea that conceptually nothing stops TEXT and BORDER to have a background color. It feels simpler and more explicit when defining text in a Clay layout but is kind of redundant too. Here was the question I asked: https://github.com/nicbarker/clay/issues/409

What are your thoughts on both approaches?

AlexB52: That makes perfect sense. I've overlooked text overflows 👍

Mivirl:
> Mouse coordinates will only go up to 223 for now due to a [bug in termbox2](https://github.com/termbox/termbox2/pull/101).
> 
> If the patch is applied mouse support will work with larger terminal sizes (and the git commit hash in CMakeLists.txt will need to be updated to use it).

The patch has been applied upstream, so mouse support works on all sizes now.

I've also added image support with a few modes:
- `CLAY_TB_IMAGE_MODE_PLACEHOLDER` - Draws the same placeholder as before
- `CLAY_TB_IMAGE_MODE_BG` - Just draws background colors. Fastest method
- `CLAY_TB_IMAGE_MODE_ASCII_FG` - Just draws foreground ascii characters
- `CLAY_TB_IMAGE_MODE_ASCII_FG_FAST` - Same as above, but considers fewer options
- `CLAY_TB_IMAGE_MODE_ASCII` - Draw foreground and background with ascii characters. Slowest method.
- `CLAY_TB_IMAGE_MODE_ASCII_FAST` - Same as above, but considers fewer options
- `CLAY_TB_IMAGE_MODE_UNICODE` - Draw foreground and background with unicode characters
- `CLAY_TB_IMAGE_MODE_UNICODE_FAST` - Same as above, but considers fewer options

Unicode mode:

clay_tb_image_unicode_shark

clay_tb_image_unicode_castle

clay_tb_image_unicode_flower

Ascii mode:

clay_tb_image_ascii_shark

Ascii foreground-only mode:

clay_tb_image_ascii_fg_shark

Background-only mode:

clay_tb_image_bg_shark


<details><summary>Image resource attribution</summary>
All images unchanged, but rendered though code in clay_renderer_termbox2.c

- `512px-Shark_antwerp_zoo.jpeg`
  - Retrieved from <https://commons.wikimedia.org/wiki/File:Shark_antwerp_zoo.jpg>
  - License: [Creative Commons Attribution 3.0 Unported](https://creativecommons.org/licenses/by/3.0/)
- `512px-Balmoral_Castle_30_July_2011.jpg`
  - Retrieved from <https://commons.wikimedia.org/wiki/File:Balmoral_Castle_30_July_2011.jpg>
  - License: [Creative Commons Attribution-ShareAlike 3.0 Unported](https://creativecommons.org/licenses/by-sa/3.0/)
- `512px-German_Shepherd_(aka_Alsatian_and_Alsatian_Wolf_Dog),_Deutscher_Schäferhund_(Folder_(IV)_22.jpeg`
  - Retrieved from <https://commons.wikimedia.org/wiki/File:German_Shepherd_(aka_Alsatian_and_Alsatian_Wolf_Dog),_Deutscher_Sch%C3%A4ferhund_(Folder_(IV)_22.JPG>
  - License: [Creative Commons Attribution-ShareAlike 3.0 Unported](https://creativecommons.org/licenses/by-sa/3.0/)
- `512px-Rosa_Cubana_2018-09-21_1610.jpeg`
  - Retrieved from <https://commons.wikimedia.org/wiki/File:Rosa_Cubana_2018-09-21_1610.jpg>
  - License: [Creative Commons Attribution-ShareAlike 4.0 International](https://creativecommons.org/licenses/by-sa/4.0/)
- `512px-Vorderer_Graben_10_Bamberg_20190830_001.jpeg`
  - Retrieved from <https://commons.wikimedia.org/wiki/File:Vorderer_Graben_10_Bamberg_20190830_001.jpg>
  - License: [Creative Commons Attribution-ShareAlike 4.0 International](https://creativecommons.org/licenses/by-sa/4.0/)
</details>

The rendering process is not fast for large terminal windows, so previously rendered cells are cached in the image object to be reused without needing to convert on every render. Some images may additionally be converted across multiple renders to improve responsiveness, also using the cache.

_Each image object should only be used with one width/height. If an image is used twice with two different sizes, each time the image is rendered at one size it will replace the cache and require the other size to be converted again._ This is very noticeable when the debug image preview is visible at the same time as an image that is only partially rendered, since the larger image will never finish rendering.

There's probably a better way to cache image renders that doesn't have this issue, but this was the most straightforward approach I could come up with. What might work is an array/hash of cached images could be stored with new cached versions created for every size, then have that invalidated on terminal resize so unnecessary versions aren't stored.

nicbarker:
I finally had a chance to test this locally on my machine, and it looks so amazing. Thank you so much for all the hard work here! I'm happy to merge this as is now, if you feel happy with it 🙂

nicbarker:
I did have to make one tiny change which was just bumping the `CMakeLists.txt` C version to C11 for the unicode literal support with `U'`.

PR #423, opened 2025-05-26
[CMake] Set CMake FetchContent GIT_TAG for SDL_ttf

weslord:
Tags for SDL3 releases are now available in the SDL_ttf repo


nicbarker:
Ah great, thank you I forgot to update the pointer to main. Much nicer than things breaking randomly off main 🙂

PR #425, opened 2025-05-26
[CMake] Remove redundant debug & release flag variable definitions

weslord:
At one point these commands were injecting additional flags, but now it looks like they define a variable as itself (presumably to no effect?).

Two more files also have some unusual flags, but I'm less confident that removing them won't have unexpected side-effects so I haven't touched them in this PR:
https://github.com/nicbarker/clay/blob/ce2475ba739c04938616d59e6048a1355f4fdd06/examples/terminal-example/CMakeLists.txt#L14-L15

https://github.com/nicbarker/clay/blob/ce2475ba739c04938616d59e6048a1355f4fdd06/examples/introducing-clay-video-demo/CMakeLists.txt#L29

### DIFF
--- a/examples/SDL2-video-demo/CMakeLists.txt
+++ b/examples/SDL2-video-demo/CMakeLists.txt
@@ -44,13 +44,6 @@ target_link_libraries(SDL2_video_demo PUBLIC
     SDL2_image::SDL2_image-static
 )
 
-if(MSVC)
-  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
-else()
-  set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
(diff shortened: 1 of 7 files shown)


nicbarker:
Ah, apologies these really should be commented better. I just leave them there because I sometimes want to switch the flags on locally so I change things like:
```diff
- set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG}")
+ set(CMAKE_C_FLAGS_DEBUG "${CMAKE_C_FLAGS_DEBUG} -Wall -Werror")
```

When I was experimenting a little while ago, this seemed to be the only way I could get my local cmake to accept more flags without clobbering the auto injected ones from CLion. If there's a better way I'd be happy to hear about it 🙂

weslord:
Makes sense!

> If there's a better way I'd be happy to hear about it 🙂

I’m teaching myself about CMake by reading these files, so I have no better suggestions.

PR #430, opened 2025-06-03
[Renderers/Raylib] Fix element float distortion

rootBrz:


nicbarker:
Great catch, thank you!

PR #444, opened 2025-06-22
[#443] [Text/TTF] Font sizing is ignored

### LINKED ISSUE #443: [Text/TTF] Font sizing is ignored
The SDL2 and SDL3 examples and renderer ignores the `fontSize` property of the text rendering data both on the render and the measurement side.

wolfee001:
Fixes #443


nicbarker:
Looks good to me, great work!

PR #451, opened 2025-06-24
Fix aspect ratio display, closes #449

### LINKED ISSUE #449: Aspect ratio gets casted to an Int in the debugger
Hi.

You can see [here](https://github.com/nicbarker/clay/blob/main/clay.h#L3611) that you are casting the aspect ratio to an integer when displaying it in the Debug window. This obviously rounds it, and makes it not accurate when displayed. Perhaps a "Clay__FloatToString" function should be added.

tralph3:
Tiny change, please tell me if the identifier name suits your conventions.


nicbarker:
This looks great, thank you!

PR #462, opened 2025-07-08
[bindings/odin] SizingFit and SizingGrow default parameters

A1029384756:
`SizingFit` and `SizingGrow` now take default parameters instead of having to fill with a default struct. this takes the usage from:
```odin
clay.SizingGrow({})
clay.SizingFit({})
```
to
```odin
clay.SizingGrow()
clay.SizingFit()
```
in the default case


A1029384756:
@tralph3 i considered it but decided against it because the common use case for that proc is to actually fill in the info, meanwhile `SizingFit` and `SizingGrow` are nearly always left blank

nicbarker:
Thanks for the great work here! Little improvements to the ergonomics really add up 🙂

PR #477, opened 2025-08-01
Added a new field to simplify creating and rendering custom elements.

boringlily:
I have been handling a few different custom commands and didn't like having to create a type and data wrapper before passing it into the custom data. Since in most cases a numerical based ID will be used to differentiate between types of data I added an extra field to account for that by default to simplify construction of custom elements.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -512,6 +512,7 @@ CLAY__WRAPPER_STRUCT(Clay_FloatingElementConfig);
 typedef struct Clay_CustomElementConfig {
     // A transparent pointer through which you can pass custom data to the renderer.
     // Generates CUSTOM render commands.
+    uint32_t customCommandId;
     void* customData;
 } Clay_CustomElementConfig;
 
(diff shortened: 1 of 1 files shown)


nkorth:
Hmm, rather I was the one who misunderstood what you were suggesting. Sounds like you have custom elements which only require a single pointer to **existing** data (not a render-specific struct like Clay's RenderData?) In which case I agree, it does sound annoying to bother allocating/managing a bunch of tiny wrapper structs.

boringlily:
> Note that elements have also a `.userData` field. I'm using the `.customData` field to specify the type of custom element, and then put in `.userData` relevant rendering information.

The issue isn't that there isn't a way to pass the data I want through; I just don't like the idea of passing non-pointer data through a pointer. I know I could create some globals and pass those as the "Id" values, but that just feels way tooo tacky to me. But that is a good idea you suggest.

PR #480, opened 2025-08-08
[Renderers/Sokol] Text measurement/rendering width mismatch in sokol_clay

andrewmd5:
The text bounding box width was incorrect because the measurement function was using logical pixels while the rendering function was using physical pixels (scaled by DPI). This caused a mismatch where:

- sclay_measure_text() set font size and letter spacing without DPI scaling
- sclay_render() applied DPI scaling to both font size and letter spacing
- fontstash's fonsTextBounds() returned physical pixel measurements even when given logical pixel sizes, leading to incorrect width calculations

The fix ensures consistent DPI scaling by:
1. Scaling font size and letter spacing by dpi_scale during measurement
2. Dividing the returned bounds by dpi_scale to convert back to logical pixels
3. This ensures Clay receives measurements in logical pixels that match what will actually be rendered

This was particularly noticeable with long continuous text where the discrepancy accumulated across many characters. The issue was font-size dependent because different font sizes have different rounding behaviors in the underlying font rendering system.

Also adds text alignment support to properly measure and render centered/right-aligned text.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -566,6 +566,8 @@ typedef struct Clay_TextRenderData {
     uint16_t letterSpacing;
     // The height of the bounding box for this line of text.
     uint16_t lineHeight;
+    // Specifies the alignment of the text within its container.
+    Clay_TextAlignment textAlignment;
 } Clay_TextRenderData;
 
 // Render command data when commandType == CLAY_RENDER_COMMAND_TYPE_RECTANGLE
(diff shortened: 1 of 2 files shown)


nicbarker:
Thank you for the great work here!
Just before we get this merged, would you be able to clarify for me the need to pass the text alignment to sokol here?
Clay's text render commands don't include line breaks, and the bounding box for each text render command should be tight around the unwrapped text.
The only reason I'm asking is because `Clay_TextRenderData` is part of a union that we include in every render command, and because of struct padding, adding that 1 byte enum for text alignment is unfortunately going to cost us an 8 byte increase to the size of the entire render command struct 😅

PR #484, opened 2025-08-11
Added instructions for building examples

diraneyya:
The bulk of the instructions were generated using GitHub copilot which were tested and edited manually by me.


diraneyya:
I am happy with these instructions and I won't be putting any more effort or time into them.

PR #485, opened 2025-08-11
Add scissor data stack to fix nested clip elements displaying incorrectly

### LINKED ISSUE #466: Nested clip elements behave incorrectly
When an element X has the `clip` property set and contains elements that also have the `clip` property, X fails to clip its contents properly. 

<details>
<summary>Minimal reproducible example and pictures:</summary>
<br>
With the Raylib renderer (modified from the `raylib-sidebar-scrolling-container` example):

```c
    Clay_BeginLayout();
    CLAY({
        .layout = {
            .sizing = {
                .width = CLAY_SIZING_GROW(0),
                .height = CLAY_SIZING_GROW(0),
            },
            .padding = { 16, 16, 16, 16 },
            .childGap = 16,
            .layoutDirection = CLAY_TOP_TO_BOTTOM,
            .childAlignment = {
                .y = CLAY_ALIGN_X_CENTER,
            },
        },
        .backgroundColor = { 0, 0, 0, 255 },
    }) {
        CLAY({
            .layout = {
                .sizing = {
                    .width = CLAY_SIZING_GROW(0),
                    .height = CLAY_SIZING_PERCENT(0.5),
                },
                .padding = { 16, 16, 16, 16 },
                .childGap = 16,
                .layoutDirection = CLAY_TOP_TO_BOTTOM,
            },
            .clip = {
                .vertical = true,
                .childOffset = Clay_GetScrollOffset(),
            },
            .backgroundColor = {50, 50, 50, 255}
        }) {
            for (size_t i = 0; i < 4; i++) {
                CLAY({
                    .clip = {
                        .horizontal = true,
                        .childOffset = Clay_GetScrollOffset(),
                    },
                    .layout = {
                        .sizing = {
                            .height = CLAY_SIZING_PERCENT(0.4),
                        },
                    },
                }) {
                    CLAY_TEXT(
                        CLAY_STRING("long text long text long text long text long text long text long text long text long text long text long text long text"),
                        CLAY_TEXT_CONFIG({
                            .fontId = 0,
                            .letterSpacing = 5,
                            .fontSize = 26,
                            .textColor = {255,255,255,255}
                        })
                    );
                }
            }
        }
    }
    return Clay_EndLayout();
```

Image

</details>

To the best of my knowledge, this behavior is a consequence of three issues:
1. When the Raylib renderer encounters two `SCISSOR_START` commands without a `SCISSOR_END` in between, it overwrites the clip bounding box from the first `SCISSOR_START` with the one from the second `SCISSOR_START`, and the `SCISSOR_END` corresponding to the second `SCISSOR_START` disables all clipping. This causes elements rendered after the first `SCISSOR_END` but before the second one to not be clipped. (I fixed this by maintaining a stack of bounding boxes inside the renderer; a `SCISSOR_START` causes the corresponding bounding box to be pushed onto the stack, and a `SCISSOR_END` pops an element off the stack and restores clipping corresponding to the bounding box that is now on the top of the stack.)
2. If a clip element A contains another clip element B, it is possible for the bounding box provided in the `SCISSOR_START` command for element B to not be contained within the bounding box provided in the `SCISSOR_START` for element A. (This caused problems for me because naively calling Raylib's `BeginScissorMode()` with the bounding box provided by the command fails to account for the encompassing clip elements. To fix this, the bounding box of a new `SCISSOR_START` command should itself be clipped according to the bounding box of the top element of the stack.)
3. When a `clip` element is off screen, neither the `SCISSOR_START` nor the `SCISSOR_END` for that element should be generated, but in reality the `SCISSOR_END` command alone is generated. This violates the invariant that a `SCISSOR_END` always has an accompanying `SCISSOR_START` (and puts the bounding box stack in an invalid state if it is implemented as described in (1)).

On a somewhat related note, the maximum number of scroll elements seems to be (somewhat arbitrarily) limited to 10.

I have my own fixes for all of these issues and I'm happy to make a PR, but I'd like to clarify some architectural questions first. Issue (3) is clearly a problem with the Clay layout algorithm itself, but I implemented fixes for (1) and (2) directly in the renderer. Is this the preferred separation of responsibilities? I can certainly see an argument for implementing the fixes directly in the layout algorithm, since the "correct" behavior seems pretty universally desirable.

ambareesh1510:
Fixes #466.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -1221,6 +1221,13 @@ typedef struct {
 
 CLAY__ARRAY_DEFINE(Clay__LayoutElementTreeRoot, Clay__LayoutElementTreeRootArray)
 
+typedef struct Clay__ScissorData {
+    Clay_BoundingBox boundingBox;
+    uint32_t id;
+} Clay__ScissorData;
+
+CLAY__ARRAY_DEFINE(Clay__ScissorData, Clay__ScissorDataArray)
+
 struct Clay_Context {
     int32_t maxElementCount;
     int32_t maxMeasureTextCacheWordCount;
@@ -1276,6 +1283,7 @@ struct Clay_Context {
     Clay__MeasuredWordArray measuredWords;
     Clay__int32_tArray measuredWordsFreeList;
     Clay__int32_tArray openClipElementStack;
+    Clay__ScissorDataArray scissorDataStack;
     Clay_ElementIdArray pointerOverIds;
     Clay__ScrollContainerDataInternalArray scrollContainerDatas;
     Clay__boolArray treeNodeVisited;
@@ -2201,6 +2209,7 @@ void Clay__InitializeEphemeralMemory(Clay_Context* context) {
     context->treeNodeVisited = Clay__boolArray_Allocate_Arena(maxElementCount, arena);
     context->treeNodeVisited.length = context->treeNodeVisited.capacity; // This array is accessed directly rather than behaving as a list
     context->openClipElementStack = Clay__int32_tArray_Allocate_Arena(maxElementCount, arena);
+    context->scissorDataStack = Clay__ScissorDataArray_Allocate_Arena(maxElementCount, arena);
     context->reusableElementIndexBuffer = Clay__int32_tArray_Allocate_Arena(maxElementCount, arena);
     context->layoutElementClipElementIds = Clay__int32_tArray_Allocate_Arena(maxElementCount, arena);
     context->dynamicStringData = Clay__charArray_Allocate_Arena(maxElementCount, arena);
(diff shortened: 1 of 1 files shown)


ambareesh1510:
Hi @nicbarker, sorry for the delay in getting to this! I've currently marked this PR as a draft because I noticed that some of the changes in #450 coincide with the changes in this PR (`closeClipElement` should be true only if the element is not offscreen -- in fact, I just copy-pasted those two lines from your PR into my code). This change is necessary for my fix to work; would you prefer that I remove it from my PR so that yours can be merged first?

nicbarker:
👋 Hello, thanks so much for the work here! I'll take a look and then let's get it merged. Totally fine to have those two lines from the other PR in here, we can close out that one 🙂

ambareesh1510:
Sounds good, I'll mark it as ready for review then.

nicbarker:
Just letting you know that I'm doing a bit of digging here, because there's a bit of trickyness around floating elements - it's possible to attach floating elements that include a clip rectangle later, and as a result we can't rely on visiting them as part of the DFS. I'm sure we can figure something out 🙂

filippocrocchini:
This PR doesn't compile in C++ here's what I changed to fix that: https://github.com/nicbarker/clay/pull/493/commits/d2f2344ffad2370ed7293a48745a01d643279f5f

PR #489, opened 2025-08-18
[Renderers/Sokol] allow usage of images with sokol renderer

rivten:
This code implements the rendering of clay images for the sokol renderer backend.

Previously there was a big todo and the code did nothing. Now it renders the image.

Several caveats :

- I’m not that smart, so I just copy pasted the code to render rectangles and adapted it to provide uv coordinates to a texture. Any bug in the rectangle rendering will be in the image rendering.
- I tried to implement corner radius but couldn’t. Maybe someone smarter than me can figure it out (I might give it a try in the future). So if any cornerRadius is specified, the image will look quite bad. Maybe we should just render a full rectangle all the time instead of a corrupted image ?
- Yet, this is strictly better than before.
- Tinting worked out of the box, which I’m quite happy with.
- I tried to give a good explanation in the comment header, but English is not my first language, so any feedback is appreciated.
- Thanks for the awesome library.

### DIFF
--- a/renderers/sokol/sokol_clay.h
+++ b/renderers/sokol/sokol_clay.h
@@ -113,6 +113,9 @@ typedef int sclay_font_t;
 typedef struct sclay_image {
     sg_image image;
     sg_sampler sampler;
+    struct {
+        float u0, v0, u1, v1;
+    } uv;
 } sclay_image;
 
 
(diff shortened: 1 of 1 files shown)


rivten:
Thanks. Well I feel like @andrewmd5 commit that he showed is better than mine.
It handles corners as well as subimages !
Give me a tiny bit of time and I’ll improve my PR with his excellent contribution :) I think it’s worth it.

rivten:
@nicbarker  I updated the pull request. I tested it a tiny bit, and it seems correct ! I’m not sure I’ve properly handled all the if-else cases in the image rendering.

I’ve extensively cherry-picked into @andrewmd5 and I feel like he should really be the one to be credited for this work ! Thanks again !!

rivten:
Updated. Indeed there was an update on sokol_gfx and sokol_gl side and I was not using the latest version. Thanks for catching that up :)

PR #492, opened 2025-08-21
Remove ID from element declaration struct and split clay macro into CLAY and CLAY_WITHID

nicbarker:
This PR removes the `.id` field from the `Clay_ElementDeclaration` struct, and instead requires that IDs be specified using a separate macro. The justification from the Discord is below:

Hi everyone, I'm just flagging a change that I have been weighing up for a while. Our handling of IDs at the moment is a little undercooked and has some edge cases, and one of the reasons is a poor API decision on my part. The ElementDeclaration struct that you pass to the `CLAY` macro currently accepts an ID as a field, which actually doesn't make sense:
ElementDeclaration structs are supposed to be reusable so you can define styles with them, and if an ID is defined there and reused, it will create a duplicate id error.

The `_ID_LOCAL` macros depend on the current open element to base their ID on, and as a result need to be used inside the layout declaration - again, doesn't make sense for a struct that is supposed to be for reusable styles.
Because you can use Clay_Hovered etc inside the element declaration struct, they resolve before the ID has actually been attached to the element.  We do some shenanigans behind the scenes to make this "magically" work, but I've seen edge cases and it's a hack, not a reasonable solution.

After wasting a couple hours messing around with macro hackery, I've come to the conclusion that the best approach will just be to remove the .id field from the element declaration struct and split the main CLAY macro in two, CLAY stays how it is, and CLAY_WITHID is the same macro but expects an ID as the first argument. Have a look at the attached images for more info: 

Screenshot_2025-08-15_at_11 09 53_am
Screenshot_2025-08-15_at_11 10 20_am

I was able to make this change in the examples with a <30 second find + replace, but I understand it's an annoying breaking change. I'm also open to other ideas if anyone can think of anything a bit more elegant!
this also has a nice side effect of saving us 32 bytes off the size of the configuration struct, which for reusable styles does add up

> Question from Thag: how would this impact the odin bindings?

good question, the way we've approached the odin bindings actually make this change much more straightforward:

Screenshot_2025-08-15_at_11 24 41_am

Screenshot_2025-08-15_at_11 24 13_am

the UI proc changes to take an optional ID param, and the id is removed from the config struct
Image
for historical context about this mistake, clay used to look like this:

```C
CLAY(
  CLAY_ID("button"),
  CLAY_RECTANGLE({ .backgroundColor = Clay_Hovered ? blue : red }),
  // etc
```

Which roughly expanded to:

```C
Clay_OpenElement();
Clay_AttachIdToOpenElement("button");
Clay_AttachRectangleToOpenElement(
  (Clay_Rectangle) { .backgroundColor = Clay_Hovered() ? blue : red }
);
```
 
which meant that `Clay_AttachIdToOpenElement()` could run before `Clay_Hovered()`
but since we changed to the declaration struct, because of function execution order:

```C
Clay_OpenElement();
Clay_ConfigureOpenElement({
  .id = CLAY_ID("button"),
  // Clay_Hovered() is called _before_ Clay_ConfigureOpenElement
  .backgroundColor = Clay_Hovered() ? blue : red
});
```
 
so we had to do some magic to make `Clay_Hovered()` correctly read the previous frame data to compute the hover, even though at this point in the hierarchy, we have no idea what the ID of the current element is, because `ConfigureOpenElement` hasn't run
bit of a chicken and egg problem haha
at the moment we handle it by attaching an auto generated id when `Clay__OpenElement` is called, then if the declaration has an ID in it, we point that second ID to the same hash map item as an "alias"

but it means that hover and other effects can fail if the order of elements within a container isn't exactly the same between frames - which creates major issues for transitions, which is why I'm fixing this now 👍
TL;DR I made a mistake and now everyone has to pay 😂


bullno1:
`CLAY` is so nice to type though I wish it were the default instead and `CLAY_NO_ID` is the compatibility.

You can even add: `Clay_ElementDeclarationWithId` which is the same as: `Clay_ElementDeclaration` but with the ID field and something like: `CLAY_LEGACY` can use that.

nicbarker:
> `CLAY` is so nice to type though I wish it were the default instead and `CLAY_NO_ID` is the compatibility.
> 
> You can even add: `Clay_ElementDeclarationWithId` which is the same as: `Clay_ElementDeclaration` but with the ID field and something like: `CLAY_LEGACY` can use that.

That's an interesting point - I wonder if having the ID mandatory as the first argument is a better overall design for the default `CLAY` macro. It does incentivise having stable IDs for most elements, which I think is best anyway.

Perhaps we should run a poll in the discord?

mgubi:
What about using a CLAY macro with one or two parameters? E.g. with solutions from [this](https://stackoverflow.com/questions/3046889/optional-parameters-with-c-macros) stack exchange discussion? I guess should look something like (my own version, possibly wrong)
```
#define CLAY_WITH_ID(id, s) ... // add code for element instantiation with id
#define CLAY_NO_ID(s) ... // add code for element instantiation without id
#define CLAY_CHOOSER(a, b, c, ...) c
#define CLAY(...) CLAY_CHOOSER(__VA_ARGS__ , CLAY_WITH_ID, CLAY_NO_ID)(__VA__ARGS__)
```
which does not look worser than the macro stuff Clay already has, and would allow to use a simple notation for the main entrypoint of the library.

nicbarker:
@dananderson I agree that the standard API is a little verbose, but luckily with C you can just do a simple `#define CLAYA CLAY_AUTO_ID` anywhere in your own code and rename it how you like 🙂

Generally I like to err on the side of naming being more clear by default than more concise, but I can totally understand that it gets cumbersome when you have to type it over and over!

PR #498, opened 2025-08-30
clay.h: Fix struct with default initializer not using CLAY__DEFAULT_STRUCT

victor-Lopez25:
clay.h: Fix struct with default initializer not using CLAY__DEFAULT_STRUCT
no other changes


victor-Lopez25:
I also fixed a lot of warnings with type conversions locally, if you want that I'll also add it to the PR.

PR #505, opened 2025-09-08
fix: unused extra macro param

fgungor:
Build fix for CLAY_ID_LOCAL() usage


nicbarker:
Hello, and thanks for catching this!

PR #510, opened 2025-09-16
Fix the button events in the SDL example

pmasschelier:
Short fix for the SDL3 example clicks


plungepool:
This fixed a problem I was having with the example 👍 

Clicking the buttons did not update the UI to select them until another mouse event was triggered, including clicking the button again or even just moving the cursor. Potentially this is a deeper bug or unintended behavior elsewhere? But this fix does make the demo work as it seems like was intended.

PR #521, opened 2025-09-27
Fix `Clay__GetParentElementId` for new id scheme

bullno1:
Because `Clay__GetParentElementId` is now evaluated before `Clay__OpenElementWithId`, the parent is now actually the current open element.
Going to the "grandparent" as it is now will result in an id clash


nicbarker:
To prepare for the transition branch, I'm going to merge this in now and rebase it. Although as you mentioned, this function is going to be renamed to `Clay_GetOpenLayoutElementId()` and moved to the public API, as it's useful for getting the ID of `CLAY_AUTO_ID` elements 🙂

PR #523, opened 2025-09-30
Fixed termbox2 demo build, added scroll functionality

Godje:
## Fixing the demo
- Demo would not compile due to failure to check out the commit tags (they might have been squashed away or pruned).
- Demo would crash instantly due to a "duplicate ID" issue. I statically assigned an ID to the component that was causing this trouble (termbox_settings).

## Added horizontal scroll functionality
The termbox2 renderer and the demo **supported** scroll functionality previously, but they were never *shown* in the demo. I added horizontal scrolling for the user to be able to see the rest of the palette test window through `.clip` exactly as it is done in the documentation.

## Things that will stay "broken"
Due to how string slices are rendered, horizontal scrolling will trim the string in the wrong direction. It seems that the only way to mitigate this without touching the renderer is to compute strings with the proper trimmed length, but computed strings would require allocating them outside of the layout build. Implementation of this would bloat the demo code, so I kept this out.

### DIFF
--- a/examples/termbox2-demo/CMakeLists.txt
+++ b/examples/termbox2-demo/CMakeLists.txt
@@ -8,7 +8,7 @@ set(FETCHCONTENT_QUIET FALSE)
 FetchContent_Declare(
     termbox2
     GIT_REPOSITORY "https://github.com/termbox/termbox2.git"
-    GIT_TAG "9c9281a9a4c971a2be57f8645e828ec99fd555e8"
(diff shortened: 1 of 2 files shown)


nicbarker:
Thanks for fixing this one up! I just merged another commit that fixed the base issue that was causing the duplicate ID, but we can leave in the explicit ID cause I think it's a good ID to use explicit IDs where possible anyway 🙂

PR #524, opened 2025-10-01
Fixing termbox2-image-demo build error

Godje:


nicbarker:
Thank you for fixing these up!

PR #525, opened 2025-10-03
[Renderers/termbox] fixed horizontal text culling bug

Godje:
##### note to Maintainer
This branch includes the code from PR #523 . If this merges first, the other #523  can be closed.

# Bug (before)
Incorrect trimming behavior appears when the text's `boundingBox.x` coordinate starts going below negative numbers. Instead of culling the text from the left, it trims it from the right.

### Before scrolling

image

### After scrolling

image

# Fix (after)
Added a small loop before the main text rendering loop that increments the "text printed counter" (`i`) without actually printing the text, which then makes the main renderer loop start from the correct character when printing.

### Before scrolling

image

### After scrolling
image


nicbarker:
Thanks for the great work here!

PR #533, opened 2025-10-05
Fixed incorrect assignment in README example

IsaiahPaget:
Just noticed that the README and the example differ in how you assign an id to a clay.UI in the Odin bindings.

### DIFF
--- a/bindings/odin/README.md
+++ b/bindings/odin/README.md
@@ -21,7 +21,7 @@ CLAY({ .id = CLAY_ID("Outer"), .layout = { .padding = CLAY_PADDING_ALL(16) } }) 
 
 ```Odin
 // Odin form of element macros
-if clay.UI()({ id = clay.ID("Outer"), layout = { padding = clay.PaddingAll(16) }}) {
+if clay.UI(clay.ID("Outer"))({ layout = { padding = clay.PaddingAll(16) }}) {
     // Child elements here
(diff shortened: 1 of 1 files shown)


vegasword:
Actually the previous example was right. Here's the current definition of UI :
```odin
@(deferred_none = _CloseElement)
UI :: proc() -> proc (config: ElementDeclaration) -> bool {
	_OpenElement()
	return ConfigureOpenElement
}
```
Notice that there is no argument at all.

IsaiahPaget:
@vegasword Looks like you might have pulled an old commit because this is the code as of today.

```odin
@(deferred_none = _CloseElement)
UI_WithId :: proc(id: ElementId) -> proc(config: ElementDeclaration) -> bool {
	_OpenElementWithId(id)
	return ConfigureOpenElement
}

@(deferred_none = _CloseElement)
UI_AutoId :: proc() -> proc(config: ElementDeclaration) -> bool {
	_OpenElement()
	return ConfigureOpenElement
}

UI :: proc {
	UI_WithId,
	UI_AutoId,
}
```

UI :: proc {... is how you do function overloading in Odin.

vegasword:
Oh I see mb. Indeed I've downloaded the lastest release zip, not the lastest commit

IsaiahPaget:
All good!

PR #537, opened 2025-10-10
[Examples/official-website] updated paths in build.sh

CrackedPixel:
Fixes issues when running example site


nicbarker:
Hello, sorry for the delay in getting to this. Thanks for fixing these up!

Just for context, the reason I had these setup this way is because the example site is on the `/clay` path on the [actual website](https://nicbarker.com/clay), but I can just change my own build script on my end 🙂

PR #541, opened 2025-10-18
[Bindings/Odin] Improve example text measurement function

rats159:
This PR changes the example Odin text measurement function to
1. have proper UTF-8 support using odin's `rune` type and iterating the string by codepoint, rather than by byte
2. support clay's letter spacing configuration
The implementation was heavily inspired by raylib's own `MeasureTextEx` function


nicbarker:
Awesome, thank you for the hard work here!

Issue #73, opened 2024-12-21
[Help Wanted] Having an issue setting up Raylib and C++


Susseratal:
The title is pretty self explanatory, but I'm trying to use clay_renderer_raylib.c with C++, and whenever I compile, I get a bunch of errors which read 
``Error	C4576	a parenthesized type followed by an initializer list is a non-standard explicit type conversion syntax``
They're all pointing to different lines in clay_render_raylib.c. I'm not sure if I need to define or flag something to stop this from being an issue, but I can't figure it out. I have my project configured to ``ISO C++ 20``, and ``ISO C17`` for C++ and C respectively. 

If you need more info, let me know.

nicbarker:
I'm reasonably certain I know what the issue is here. Designated initializer syntax is slightly different between C99 and C++20. In the `clay.h` file we use the `CLAY__INIT` macro to solve this:

```
// C99 Style
(Rectangle) { .width = 5, .height = 5 };
// C++20 Style
Rectangle { .width = 5, .height = 5 };
// CLAY__INIT compatibility layer
CLAY__INIT(Rectangle) { .width = 5, .height = 5 };
```

You can see that there are a number of C99 style designated inits in the raylib renderer that aren't wrapped in `CLAY__INIT`. It's just a matter of getting those fixed up 🙂

Deltt:
Any specifics here, for beginners to get started with Clay who are having this issue?

PR #554, opened 2025-11-29
Fix clay_renderer_raylib.c errors when compiling as C++ (#73)

lucas-inacio:
Wrap designated initializers and compound literals with CLAY__INIT.

Explicity cast imageTexture.width/height to float in call to DrawTexturePro.


vdemcak:
Hey, are there any plans on merging this? I've wanted to use Clay in C++, but I'm blocked by these errors and using `-Wc++11-narrowing` is not really optimal.

PR #555, opened 2025-11-29
Right to left layout

mattahj:
I had a go at implementing right to left layout and it was surprisingly quick to get something working just from poking around with the `nextChildOffset` and a few other tweaks. If this approach is not good I would love some pointers about what to do instead.

Screenshot 2025-11-29 at 18 16 44

I have not yet tested very much or checked any edge cases, but I thought I will throw up this draft PR to check if this change would be wanted in clay? I would love to see clay used for some AAA game UIs but I think one of the blockers for that at the moment is lack of support for right to left layout for elements and text (every game I have worked on in that space needed Arabic localisation for example). I will have a go at the right-to-left text layout as well this weekend if I get some time!

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -3547,7 +3547,7 @@ void Clay__RenderDebugView(void) {
                     // .layoutDirection
                     CLAY_TEXT(CLAY_STRING("Layout Direction"), infoTitleConfig);
                     Clay_LayoutConfig *layoutConfig = selectedItem->layoutElement->layoutConfig;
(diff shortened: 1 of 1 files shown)


mattahj:
Bug: floating attached children don't behave properly. I'm guessing that where I add the `childRTLOffset` to the `childPosition` is either not correct, or I just also need to apply the same offset on the floating children

Screenshot 2025-11-29 at 18 55 10

PR #565, opened 2025-12-13
Gles3 renderer

luke10x:
GLES3 Renderer Video Demo
======================

This PR  contains a work-in-progress GLES3 renderer.
The renderer is already functional and demonstrates the core rendering pipeline.

Current features

-	Supports all Clay draw commands (Except custom commands).
-	All quad-based commands (Rectangle, Image, Border) are rendered in a single draw call.
-	All glyphs are rendered in one instanced draw call.
-	When clipping (scissoring) is used:
-	The renderer flushes draw calls before and after each scissor region.
-	Supports up to 4 fonts and 4 image textures (images can be used as atlases as custom userData of the image allows to specify UV coords), so the image texturese can be used as atlases where each sprite shares a frament of the same OpenGl texture.
-	Uses stb_image.h and stb_truetype.h as single-header dependencies for asset loading (Although loaders are swappable but STB example implementation is provided).

Currently builds on:
-	CMake / macOS, possibly other platfroms
-	Emscripten

I added two examples: one uses SDL2, and the other uses GLFW but the renderer is framework agnostic as long as Ogles is loaded.


luke10x:
> I can have a look at creating a CMake target for the renderer, which should work fine with emscripten through `emcmake`.
> 
> A CMake build can use FetchContent to grab the stb dependencies from their repositories too, which it looks like @luke10x has started.

I would appreciate looking into, and advising how to move forward. Yet I am in doubt myself, it looks like I have a working solution right now where stb dependency is managed by CMake. If this renderer gets ever merged to Clay, it does not mean that Clay would depend on stb, because all stb references are extracted to a swappable  [STB Loader](./renderers/GLES3/clay_renderer_gles3_loader_stb.c), so if some Clay user does not want or cannot use STB, theres is still the way to use renderer with alternative loader for fonts and images.

This renderer comes with 3 companion examples: 2 using SDL2, and 1 using GLFW, native builds require those libraries to build the examples. Emscripten SDK already contains both SDL2 and GLFW3, so it is not required to build them from sources. This is where different platforms would require different dependencies. For example, say we also want to have linux build, then it will also require bunch of more dependencies to load OpenGL ES functions: mesa, gl3w or glew, etc.

I would suggest that examples could have just platform specific handcrafted simple Makefiles and build instructions defined in their README files. Because the examples are not really critical for the renderer itself. I personally don't really feel any urge to to have automated build for the example code, but please direct me if you have a different opinion.

Now few words about current state: I am testing the renderer with my real-life project too, I managed to find may bugs, but I pushed fixes, and I believe the code could stay more stable from this moment on, but if you want look it into my branch, maybe you could please give me another couple of days to make sure there are no essential flaws. Also please provide feedback if you see any changes in the code that I should apply to fit Clay project better.

Thank you for looking into it.

I believe I should create an issue related to this PR where more detailed discussion can take place.

renderers/GLES3/clay_renderer_gles3.h
```diff
@@ -0,0 +1,814 @@
...
+#if defined(__EMSCRIPTEN__)
+    #include <emscripten.h>
+    #include <emscripten/html5.h>
+    #include <GLES3/gl3.h>
+    #define GLSL_VERSION "#version 300 es"
+#else
+    // Only apple computers now sorry
+    // That means it is not really GLES3 but desktop OpenGL 3
+    // Luckily, it is compatible with GLES3
+    #include <OpenGL/gl3.h>
```
flipfloppy1: CMake generates build files just fine for me on Linux, but my OpenGL header files are at `GL/*.h` or `GLES3/*.h`.

I had to change the OpenGL include here to `#include <GLES3/gl3.h>` and then the GLFW example built properly.

It would be good to detect the default header file locations for different platforms and potentially add an override for people that want to specify it themselves, using preproc macros.

nicbarker:
Just tested everything on my local machine and it looks great. Thank you for your hard work!

PR #576, opened 2026-01-25
Add Java rewrite to the 'Other implementations' section in the README

Lauriichan:
I made a reimplementation of Clay in Java, sometimes matching the original C code as closely as possible sometimes not but mostly functional.
I'm not super talented so this might not be the best possible reimplementation there is but so far it seems to be mostly working and I hope others might find it helpful either as a reference for reimplementation of maybe it is of use for them which is why I created this PR.

Note: I made a PR (#574) already but closed it due to this not being a binding so it shouldn't be in the bindings folder instead I made this new PR which made it just be a mention on the main README


Lauriichan:
Is there any chance that this will possibly be merged into the README or rather not since there hasn't been a reply to this?

Issue #471, opened 2025-07-23
Access ephemeral data from previous frame through layoutElement pointer in Clay_GetScrollContainerData


PiggybankStudios:
I've spent a while trying to track down a memory access issue that happens when I call `Clay_GetScrollContainerData` on a container BEFORE that container is created during the frame. The problem is caused by the persistent `Clay__ScrollContainerDataInternal` struct holding a `layoutElement` pointer which points to an item that was on an Ephemeral array. This `layoutElement` pointer is passed to `Clay__FindElementConfigWithType(scrollContainerData->layoutElement, CLAY__ELEMENT_CONFIG_TYPE_CLIP)` inside `Clay_GetScrollContainerData` to get a `Clay_ClipElementConfig *` which ultimately just gets passed back to the application inside `Clay_ScrollContainerData.config`. Since the Ephemeral arrays are reset at the beginning of the frame, this usage of the `layoutElement` pointer is invalid and could potentially point to anything.

Code in Question:
```cpp
Clay_ScrollContainerData Clay_GetScrollContainerData(Clay_ElementId id) {
    Clay_Context* context = Clay_GetCurrentContext();
    for (int32_t i = 0; i < context->scrollContainerDatas.length; ++i) {
        Clay__ScrollContainerDataInternal *scrollContainerData = Clay__ScrollContainerDataInternalArray_Get(&context->scrollContainerDatas, i);
        if (scrollContainerData->elementId == id.id) {
            Clay_ClipElementConfig *clipElementConfig = Clay__FindElementConfigWithType(scrollContainerData->layoutElement, CLAY__ELEMENT_CONFIG_TYPE_CLIP).clipElementConfig; // <-- Using layoutElement here is unsafe!
            if (!clipElementConfig) { // This can happen on the first frame before a scroll container is declared
                return CLAY__INIT(Clay_ScrollContainerData) CLAY__DEFAULT_STRUCT;
            }
            return CLAY__INIT(Clay_ScrollContainerData) {
                .scrollPosition = &scrollContainerData->scrollPosition,
                .scrollContainerDimensions = { scrollContainerData->boundingBox.width, scrollContainerData->boundingBox.height },
                .contentDimensions = scrollContainerData->contentSize,
                .config = *clipElementConfig, // <-- clipElementConfig is just being passed back to the caller
                .found = true
            };
        }
    }
    return CLAY__INIT(Clay_ScrollContainerData) CLAY__DEFAULT_STRUCT;
}
```

[...]

PR #584, opened 2026-03-19
Fix Issue #471: Access ephemeral data from previous frame through layoutElement pointer in Clay_GetScrollContainerData

lorlouis:
The relevant context is in issue #471, the fix proposed by @dhanton is now possible with the new way IDs are assigned to Clay elements.


nicbarker:
Hello, thanks for the work here! This issue was fixed in another PR: https://github.com/nicbarker/clay/pull/606/changes
So we can close this one 🙂

PR #592, opened 2026-03-31
[bindings/odin] add transitions api

A1029384756:
this will remain as a draft until the library decides what to do about tinted images. ~~right now, there is a bug where extra render commands are generated, producing images that are covered by colored rectangles~~. this is just due to a change in how background color and images work together. as it stands, the background color on the image data now will produce and extra render command and should probably be removed from the render command struct


nicbarker:
Just ran this locally, looks great! Thanks for the good work 🏄‍♂️

PR #598, opened 2026-04-08
Implement `CLAY_TRANSITION_PROPERTY_CORNER_RADIUS`

ZeroMemes:
Support for `CLAY_TRANSITION_PROPERTY_CORNER_RADIUS` was not implemented by #579. Not sure if this was an intentional decision pending additional considerations or just forgotten, but I couldn't find any comments or commits that explicitly mentioned it.

As a side note, I've chosen to use `Clay__FloatEqual` when evaluating the active properties, which is what the `X/Y/WIDTH/HEIGHT` properties use, but color properties are also backed by floating point values and use `Clay__MemCmp`.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -563,6 +563,7 @@ typedef struct {
     Clay_BoundingBox boundingBox;
     Clay_Color backgroundColor;
     Clay_Color overlayColor;
+    Clay_CornerRadius cornerRadius;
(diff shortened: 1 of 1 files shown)


nicbarker:
Thanks for this, it was literally just an oversight 😅

PR #599, opened 2026-04-08
[Core] Add text elision

develOseven:
Should elide like in Qt QML.


nicbarker:
This is a really awesome contribution! Thank you for the work. I'll try to get to this over the next few days.

PR #600, opened 2026-04-09
Fix #189

### LINKED ISSUE #189: `CLAY__STRING_LENGTH` internal macro uses incorrect formula
`CLAY__STRING_LENGTH` is defined as:
```c
#define CLAY__STRING_LENGTH(s) ((sizeof(s) / sizeof((s)[0])) - sizeof((s)[0]))
```
However, this is incorrect. It technically works under most circumstances, which is why it's gone unnoticed, however, this returns the wrong length for non-UTF-8 strings. Looking at the corrected macro makes it clear why:
```c
#define CLAY__STRING_LENGTH(s) ((sizeof(s) / sizeof((s)[0])) - 1)
```
Instead of simply subtracting the teminating null after calculating the array length, the current definition overthinks things and instead subtracts the size of the characters in bytes.

While this admittedly isn't a very critical bug, it's still a silly one, it stil *actually cropped up in my environment somehow*, and it is a one-line fix.

cD1rtX3:
Fixes #189.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -95,7 +95,7 @@
 
 #define CLAY_SIDI_LOCAL(label, index) Clay__HashStringWithOffset(label, index, Clay_GetOpenElementId())
 
-#define CLAY__STRING_LENGTH(s) ((sizeof(s) / sizeof((s)[0])) - sizeof((s)[0]))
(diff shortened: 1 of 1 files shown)


cD1rtX3:
Superseded by #614.

PR #601, opened 2026-04-10
[Core] Fix missing text element guards / incorrect union access

bullno1:
There are several places where Clay will use `Clay_LayoutElement.config.layout.sizing` to make layout decisions.
However, this will have an undetermined value for a text element due to being overlapped with an union.

This PR introduces a special internal `CLAY__SIZING_TYPE_TEXT` so that most of the layout logic can be kept intact without having to do an explicit check for whether the element is a text element.

### DIFF
--- a/clay.h
+++ b/clay.h
@@ -306,8 +306,6 @@ typedef CLAY_PACKED_ENUM {
     CLAY__SIZING_TYPE_PERCENT,
     // Clamps the axis size to an exact size in pixels.
     CLAY__SIZING_TYPE_FIXED,
-    // For internal use
-    CLAY__SIZING_TYPE_TEXT,
 } Clay__SizingType;
 
 // Controls how child elements are aligned on each axis.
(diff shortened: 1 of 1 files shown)


nicbarker:
This is a great catch, during the refactor I completely missed my previous assumption that text elements also had a valid layout config.

nicbarker:
AFAIK, the old text layout config was just the empty default - I think we might be able to avoid making a new sizing mode and just return the default. I've pushed the change, but please let me know if you think it isn't correct here.

nicbarker:
Found a few more cases where we weren't guarding correctly against text elements, fixing them up.

bullno1:
> AFAIK, the old text layout config was just the empty default - I think we might be able to avoid making a new sizing mode and just return the default. I've pushed the change, but please let me know if you think it isn't correct here.

Yeah, it becomes FIT which I guess is safe enough because text can't have children.

I only added a new type because I didn't want to dive into the details.

PR #602, opened 2026-04-10
Hide irrelevant data from the inspector for text element

bullno1:
Since a text element can't have layout config and the data overlaps the text config in an union, it's just misleading junk and should not be displayed.


nicbarker:
LGTM, thanks for the great work!

PR #603, opened 2026-04-11
Add userdata support to transition

bullno1:
It's relatively simple. The `userData` field in the transition config will be passed as target.
It's up to the user to decide the allocation and lifetime policy.

Personally, I just use arena and it's copied repeatedly during exit: https://github.com/bullno1/bgame/blob/7d11795ec1ddb73fc47f736bc7b4ee51061117a0/src/ui.c#L368-L370

This allows arbitrary customization on the user side.


cowboyd:
Is there any reason not to have this? We're integrating from JavaScript, and would be very handy to support custom transitions 😁

PR #606, opened 2026-04-12
Fix Clay_GetScrollOffset not finding the same element accross frames

bullno1:
If there is a new element between frames, the order of allocation is not the same so pointer comparison doesn't work.
The comparison has to be done with id instead.

This can manifest as a flashing of the scroll container since it will reset to 0 for a single frame.
This can be observed even in the builtin inspector.


bullno1:
There are  other parts in the code where the same pointer comparison is done but it happens in the same frame so I guess it's fine.

However, the pointer does make the struct bigger which uses a bit more memory.

nicbarker:
Great catch!

PR #607, opened 2026-04-12
Links for D bindings

cyrusmsk:
Added links for D bindings

clayd - is a clay D bindings with the same functionality.
Currently it has only SDL3 renderer. And simple example ready to run (tested on linux and macos)

clayui - is the UI component library, based on clayd. Has raylib and SDL code inside


nicbarker:
Thanks for the work here!

PR #608, opened 2026-04-12
[Core] Fix floating element exit transitions

bullno1:
For some reason, exiting floating elements will increase the parent's size when reattached.

This fixes it but I'm not really sure it's correct.


bullno1:
This seems to mess up z-order

Edit: should be fixed

nicbarker:
Thanks for the great work! I modified the code that converts the element to floating when parent exits so that we don't need the ternary for the `parentId`.

PR #650, opened 2026-08-24
Add support for multibyte UTF-8 characters in Raylib_MeasureText

hensing1:
The present implementation for the text measuring function in the raylib renderer does not support multibyte UTF-8 codepoints, such as German umlauts.

Worse, since the leading bit of the first byte in a multibyte UTF-8 sequence is always a 1, the index calculated by `int index = text.chars[i] - 32;` is guaranteed to be negative, making `fontToUse.glyphs[index].advanceX` read into random memory.
Also, the order in which the glyphs are present in the `glyphs` field of a raylib font can be arbitrary, for example when loaded with custom code points using raylib's `LoadFontEx()`.

This fix employs the functions `GetCodepoint()` and `GetGlyphInfo()`, both provided by raylib.

It also addresses the point raised in #634 .


rats159:
I did something similar for the Odin raylib renderer, and I recommend not making this the default, a lot of text is so ascii, and it's significantly faster to work with than full utf-8. In the Odin version I have measure_text_ascii and measure_text_utf8, with measure_text being an alias for the ascii one. I'd recommend a similar approach here to keep performance, while adding the more expensive option when necessary

hensing1:
@rats159 fair enough, I've added the old function again. Lmk if this is ok

